PR #4, opened 2021-06-25
Added some convenience to the Error effect

mmhat:
This PR adds type aliases of `Error` effect for common exception types, namely `IOError` and `SomeException`. In addition, specialized versions of `catchError` and `tryError` are provided by that module.

### DIFF
--- a/src/Effectful/Error.hs
+++ b/src/Effectful/Error.hs
@@ -1,10 +1,16 @@
 -- | Support for checked exceptions.
 module Effectful.Error
  ( Error
+ , ErrorIO
+ , ErrorSome
  , runError
  , throwError
  , catchError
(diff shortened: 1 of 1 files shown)


arybczak:
Thanks for the PR, but unfortunately this isn't right. The `Error` effect is for catching errors of a particular type only, the fact that it uses exceptions underneath is an implementation detail.

The Error effect is really "ExceptT done right", not a general mechanism for handling exceptions. That's what functions from the `exceptions` library are for.

mmhat:
Oh, ok! I missed that part. Thank you for clarification. I will open another PR and contribute some documentation on that.

mmhat:
BTW: Thank you for this libary! It is really fun working with it :-)

PR #7, opened 2021-06-25
Pass CallStack to error handlers

### LINKED ISSUE #6: Pass CallStack to error handlers instead of [String]
Right now the functions for exception handling expose the call stack of the location where the exception occurred as a list of strings.
IMHO it would be better to expose the `CallStack` type directly and not the pretty-printed list of call sites.
What is the reason for that design decision in the first place?

mmhat:
Fixes #6


arybczak:
Thanks!

PR #8, opened 2021-06-25
Doctest test suite

mmhat:
This PR adds a new test suite verifying the examples in the documentation of the source code using [doctest](https://github.com/sol/doctest#readme).


arybczak:
This is not needed, `haskell-ci` supports doctest and makes CI run it out of the box (see e.g. #10 that failed because of invalid doctest).

mmhat:
Oh, cool, I didn't know that!

PR #10, opened 2021-06-25
Documentation: Clarify that Error effect basically is ExceptT

mmhat:


src/Effectful/Error.hs
```diff
@@ -1,4 +1,20 @@
...
+   `Control.Monad.Except.ExceptT` monad transformer found in the mtl library.
+   It is __not__ intended to be a general mechanism for catching errors like
+   the functions of the 'exceptions' package. For example, if you want to
+   a `Control.Exception.SomeException` you could rely on `Control.Monad.catch`:
+
+   >>> :{
+     let boom :: Eff '[IOE] String
+         boom = error "BOOM!"
+     in
+     runIOE $ boom `catch` \(e :: SomeException) ->
```
arybczak, line 13: Catching all exceptions is an antipattern, it should catch `ErrorCall`.

src/Effectful/Error.hs
```diff
@@ -1,4 +1,20 @@
--- | Support for checked exceptions.
+{- | Support for checked exceptions.
+
+   The `Error` effect provided by this module is a suitable replacement for the
+   `Control.Monad.Except.ExceptT` monad transformer found in the mtl library.
+   It is __not__ intended to be a general mechanism for catching errors like
+   the functions of the 'exceptions' package. For example, if you want to
+   a `Control.Exception.SomeException` you could rely on `Control.Monad.catch`:
```
arybczak, line 7: I think you accidentally a word ;)

src/Effectful/Error.hs
mmhat: I agree that it is good to be as specific as you can be when it comes to exception handling. However, regarding the intention of this simple example I used `SomeException` to explain that we catch an error regardless of its nature, i.e. without knowing which kind of error `boom` throws. When I was writing this I first used `readFile "does-not-exist.txt"` but I thought the use of `error` might be even more clear.

TLDR: There are two options:

1. Rewrite the example to something that justifies `SomeException`; e.g. a resource cleanup scenario.
2. Change `SomeException` to `ErrorCall` as you suggested.

I'm ok with both.

src/Effectful/Error.hs
mmhat: I'm afraid I don't quite understand your comment ;-)

mmhat:
Hm, that's weird: When I run doctest locally it passed...

src/Effectful/Error.hs
mmhat: I went with option 2.; It is the simpler one.

src/Effectful/Error.hs
arybczak: You wrote "if you want to a `Control.Exception.SomeException`", there's a missing "catch" there :slightly_smiling_face:

mmhat: :D Well, forgot to catch that exception... Fixed.

mmhat:
As far as I am concerned this can be merged.

PR #11, opened 2021-06-26
Stack configuration

mmhat:
This PR adds an initial Stack configuration with resolver LTS 18.0.


arybczak:
I don't use stack and am not familiar with how its config works. Is the `stack.yaml` file actually needed for building with stack?

mmhat:
Yes, you can simply run `stack init` in the cloned repository and it will give you a Stack configuration. But when I did this I ended up with the nightly snapshot which uses GHC 9.0.1 and in the consequence I ran into [this](https://github.com/arybczak/effectful/blob/120ac9d1548e7600e240e6a32af2af7ccc4b4b6a/src/Effectful/Writer.hs#L44) issue.
Therefore I figured it might be good to provide a configuration in order to get a better developer experience out of the box.
It's really a minor issue though...

BTW: Which GHC issue are your referring to? Can you provide a link?

arybczak:
It was most likely [this](https://gitlab.haskell.org/ghc/ghc/-/issues/19700) as the code compiled fine in the current ghc-9.2 and ghc-9.0 branch of GHC, but didn't on 9.2.0-alpha2:

```
[15 of 16] Compiling Effectful.Writer ( src/Effectful/Writer.hs, /home/unknown/Programowanie/haskell/simple-eff/dist-newstyle/build/x86_64-linux/ghc-9.2.0.20210422/effectful-0.0.0.0/build/Effectful/Writer.o, /home/unknown/Programowanie/haskell/simple-eff/dist-newstyle/build/x86_64-linux/ghc-9.2.0.20210422/effectful-0.0.0.0/build/Effectful/Writer.dyn_o )

<no location info>: error:
    ghc: panic! (the 'impossible' happened)
  (GHC version 9.2.0.20210422:
        CoreToStg.myCollectArgs
  ((case unsafeEqualityProof @(*) @Any @(IdE (Writer w_sfg1)) of
    { UnsafeRefl co_a3dZ ->
    ipv15_sgpy
    `cast` (Sub (Sym co_a3dZ) :: Any ~R# IdE (Writer w_sfg1))
    })
   `cast` (N:IdE[0] <Writer w_sfg1>_R
           :: IdE (Writer w_sfg1)
              ~R# (forall (m :: * -> *) r. Writer w_sfg1 m r)))
    @Any @Any
  Call stack:
      CallStack (from HasCallStack):
        callStackDoc, called at compiler/GHC/Utils/Panic.hs:181:37 in ghc:GHC.Utils.Panic
        pprPanic, called at compiler/GHC/CoreToStg.hs:978:33 in ghc:GHC.CoreToStg

Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug
```

However, looks like this bug is no longer triggered with `master` :facepalm: I removed the CPP. I suppose we don't need the freeze file for stack anymore then?

mmhat:
Well, if you mean the `stack.yaml.lock` file: I included it simply because it is good to do so. Have a look at [this](https://docs.haskellstack.org/en/stable/lock_files/) part of the documentation. On the other hand, it is entirely optional; Do you want me to remove it?

Regarding the CPP removal: I do think we should put a warning sign in the documentation of the `listen` function. Just to be sure that we won't receive bug reports which will be closed with "won't fix" anyway.

arybczak:
Adding `.stack-work` to `.gitignore` is fine, but `stack.yaml` isn't needed.

> I do think we should put a warning sign in the documentation of the listen function.

The bug no longer triggers (and if it does again, CI will complain and then it can be put there).

mmhat:
Well, the idea of adding the `stack.yaml` is that the developer does not have to pick a snapshot :-) But we could as well add it to the `.gitignore` such that if someone was using Stack for development it does not get in the way when running Git operation. I'm fine with that, too.

arybczak:
Thanks :)

PR #12, opened 2021-07-01
.gitignore fix for Stack configurations

mmhat:


mmhat:
I messed this up in #11 . Sorry for that noise!

PR #14, opened 2021-07-01
Benchmarks for EvEff

### LINKED ISSUE #13: Add EvEff to benchmarks
[eveff](https://hackage.haskell.org/package/eveff) is a another (recent) addition to the effect system zoo. Judging from the related paper the performance seems to be quite good, hence it would be interesting to compare it to this library.

mmhat:
Fixes #13


mmhat:
I am not planning to work on this any longer. Closing.

arybczak:
Sure. AFAIK adding `filesizes` is impossible right now because `EvEff` doesn't support IO.

PR #15, opened 2021-07-04
Added Cabal flags for benchmark suite

mmhat:
There is one flag for each effect library. Those flag control whether
that benchmark is included or not.


arybczak:
Why do you want such granularity? Wouldn't one flag for all 3 libraries be enough?

mmhat:
I used those flags to compare effectful to X one by one. But I guess one flag is sufficient.
BTW: Should we run the other benchmarks by default or not?

arybczak:
> BTW: Should we run the other benchmarks by default or not?

Yes.

Please don't gate `mtl` behind a flag, it's unnecessary as it's a GHC boot dependency.

Actually, what is the advantage of having this flag? It's not like the other effect libraries pull in tons of dependencies, they're pretty lightweight.

mmhat:
The primary motivation behind this flag is the following: When working on effectful I'd like to spot performance regression. The other benchmarks are not interesting (unless one is working on the benchmark suite itself); They just add some noise to the benchmark outputs and lead to longer development cycles. I don't care so much about the dependencies.

That's why I gated `mtl` as well.

bench/FileSizes.hs
```diff
@@ -6,7 +6,9 @@ module FileSizes where
 import Control.Exception
 import Control.Monad.IO.Class
 import Data.IORef
+#ifdef VERSION_polysemy
 import Data.Tuple
```
arybczak, line 10: AFAIK code for `eff` also uses this. You can remove this and add `eff_swap` and `poly_swap` functions below.

bench/Countdown.hs
```diff
@@ -2,7 +2,9 @@
 module Countdown where
 
 import Control.Monad.ST
+#ifdef VERSION_mtl
 import Data.Functor.Identity
```
arybczak, line 6: Please move that to the block for `mtl` imports below.

arybczak:
> When working on effectful I'd like to spot performance regression

Fair enough :+1:

bench/FileSizes.hs
mmhat: Done.

bench/Countdown.hs
mmhat: Oh, I missed that one! Done.

PR #16, opened 2021-07-04
Allow usage of unlift callbacks in other threads

mmhat:


arybczak:
How are you going to address the issue described in [here](https://github.com/arybczak/effectful/blob/453705564996c248d828cfe563c3e69f30f26e5b/src/Effectful/Internal/Monad.hs#L128)?

mmhat:
I clone (using `Effectful.Internal.Env.cloneEnv`) the environment such that each thread works on a separate copy. I followed the approach used in the `Effectful.Async` module. I am not sure if that is enough, but at least the example you refer to works fine.

arybczak:
Functions in `Async` clone the environment before the thread is spawned, AFAICS you clone it in the new thread and by then it's already too late, the example in the comment still shouldn't work unless I misunderstood something.

mmhat:
Yes and no: The clone is done after the thread is spawned, but since it is pure IO it does not know about the environment - unless you pass a reference of it to the IO action. Then we are doomed. Do I miss something?

mmhat:
> Functions in `Async` clone the environment before the thread is spawned, AFAICS you clone it in the new thread and by then it's already too late, the example in the comment still shouldn't work unless I misunderstood something.

Ah, ok, I got it! Yes, the original example fails indeed.

arybczak:
Thanks for looking into that, your code have me an idea how to solve that more or less gracefully and without performance drops. I should have it ready tomorrow.

mmhat:
Closing in favor of #17 .

PR #17, opened 2021-07-05
Add support for unlift strategies

arybczak:
Supersedes #16.


src/Effectful/Internal/Monad.hs
```diff
@@ -117,36 +127,46 @@ unsafeEff_ m = unsafeEff $ \_ -> m
...
+            pure (M.insert tid es esMap, es)
+          EQ -> pure (M.insert tid esTemplate esMap, esTemplate)
+          GT -> error $ "Number of allowed threads to run with the unlifting "
+                     ++ "function exceeded. You have to increase the argument "
+                     ++ "to BoundedConcUnlift or use UnboundedConcUnlift."
+    unEff m es
+
+-- | Lower 'Eff' operations into 'IO' ('UnboundedConcUnlift').
+unboundedConcUnliftEff :: (RunIn es IO -> IO a) -> Eff es a
+unboundedConcUnliftEff = boundedConcUnliftEff maxBound
```
mmhat, line 169: Not sure if it makes that much of a difference, but IMHO it would be better not to define `unboundedConcUnliftEff` in terms of `boundedConcUnliftEff`. Admitted, duplicating the code just to get rid of `M.size`, `compare` and the case statement might be a bit over the top but I think it is worth considering.

mmhat (commented):
I really like what you did here! :+1: Did your run some benchmarks? I am really interested in the numbers.

src/Effectful/Internal/Monad.hs
arybczak: The cost of comparing sizes is insignificant when compared to cost of cloning the environment, so it doesn't matter.

I'm actually thinking of removing `UnboundedConcUnlift` and leaving just `SeqUnlift` and `ConcUnlift Int` constructors, then setting Int to `maxBound` gives you unbounded unlift.

arybczak:
> I really like what you did here! +1

Thanks for giving me a nudge in the right direction ;) I wasn't satisfied with not being able to unlift in different threads, but I didn't have a good idea how to tackle this.

> Did your run some benchmarks? I am really interested in the numbers.

What numbers exactly do you have in mind? The current benchmarks didn't regress. In fact they're now slightly faster because `send` doesn't construct the unlifting function, effect handlers do that only when then need to. I also checked performance of `stateM` from `State.Dynamic` and it's faster now, presumably due to GHC doing better job at inlining since unlift is now constructed locally in the effect handler.

I'm unsure about two things yet:
- `withUnliftStrategy` resetting for new threads (what do you think?)
- having both `BoundedConcUnlift` and `UnboundedConcUnlift`

I'll let it sit for a bit.

src/Effectful/Internal/Monad.hs
mmhat: Yeah, this really was a nitpicking comment :-) Regarding the removal of `UnboundedConcUnlift`: I gave it a thought and I think the only reason to keep it is convenience. Note that the current implementation is in fact a bit misleading: Since we never delete `ThreadIds` (should we?) the number of calls in different threads is bound by `maxBound` anyway due to the underlying implementation using a `Data.Map`. But I guess this is mostly a theoretical problem.

mmhat:
> What numbers exactly do you have in mind? The current benchmarks didn't regress. In fact they're now slightly faster because `send` doesn't construct the unlifting function, effect handlers do that only when they need to. I also checked performance of `stateM` from `State.Dynamic` and it's faster by ~20% now, presumably due to GHC doing better job at inlining since unlift is now constructed locally in the effect handler.

These were the numbers I had in mind. Nice results! :100: 

> * `withUnliftStrategy` resetting for new threads (what do you think?)

I'll have a look.

> * having both `BoundedConcUnlift` and `UnboundedConcUnlift`

I'm not (yet) sure about this one.

mmhat:
Do you have any thoughts on garbage collecting old `ThreadId`s from `esMap`? Will that be an issue?

If so, maybe we can run something like this when the size of the map exceeds a certain threshold :
```
gc :: Map ThreadId (Env es) -> IO (Map ThreadId (Env es))
gc esMap = do
    let ks = Data.Map.Strict.keys esMap
    alive <- filterM (fmap isAlive . GHC.Conc.threadStatus) ks
    return $ Data.Map.Strict.restrictKeys esMap $ Data.Set.fromList alive
    where
        isAlive :: GHC.Conc.ThreadStatus -> Bool
        isAlive ThreadDied = False
        isAlive ThreadFinished = False
        isAlive _ = True
```
```

**Comment by arybczak** [comment]
The map is local to the scope of `liftBaseWith`/`withRunInIO`/`localUnlift` and these scopes tend to be short-lived :slightly_smiling_face: 

In theory you can stick the unlifting function in a data structure and decide to use it later, but that isn't going to work terribly well since:
- It's parametrized by `es`
- For new threads it'll give you state of the environment at the point of the unlift function's creation, not the current one, which is generally what you'd expect.

**Comment by mmhat** [comment]
I guess what I meant is:
```
foo k = withRunInIO $ \runInIO -> forever $ do
    x <- UnliftIO.Async.async $ runInIO k
    ...
```
This is of course an awfully artificial example :wink:

**Comment by arybczak** [comment]
Oh, I see. Right, it's a bit artificial, but still somewhat relevant.

There is another problem with such usage (see [here](https://hackage.haskell.org/package/base-4.15.0.0/docs/Control-Concurrent.html#t:ThreadId) for more info): since the map stores `ThreadIdS`, dead threads will not be garbage collected.

Ideally we'd like to keep track of `Weak ThreadId` made with [mkWeakThreadId](https://hackage.haskell.org/package/base-4.15.0.0/docs/Control-Concurrent.html#v:mkWeakThreadId), but `Weak` objects don't have an `Ord` instance, so we'd need to store them in an array and then lookup would take linear time :disappointed: But it wouldn't prevent garbage collection and dead threads wouldn't be a problem by design.

Linear lookup is not a problem for BoundedConcUnlift with small bound, but wouldn't work well for Unbounded if there is a lot of threads.

**Comment by arybczak** [comment]
Ok, problem solved. Now the map doesn't store direct references to ThreadIdS and EnvS and cleans itself.

File: src/Effectful/Internal/Monad.hs (hunk 1 above)
Comment (arybczak) [reply src/Effectful/Internal/Monad.hs]: Dead threads are now getting cleaned up in the unbounded case, so `bounded` and `unbounded` distinction will stay.

**Comment by arybczak** [comment]
Ah, there is one more thing.

Currently environment changes are persisted between `unlift` calls within the same thread:

```haskell
put @Int 0 -- thread-local variant of State
withRunInIO $ \unlift -> forkIO $ do
  unlift $ put @Int 1337
  ...
  unlift $ get @Int -- gives 1337
```

This is desirable in most cases (otherwise lifting e.g. `withUnmask` family of functions wouldn't work properly), but not always. E.g. in `pooledMapConcurrently` from `UnliftIO.Async` you wouldn't expect environment changes to persist between independent operations, but now they do, because worker threads are reused!

Luckily the variant that doesn't persist the environment between `unlift` calls is much easier to write than the one that does. But that will increase the number of unlift strategies.

mmhat:
Nice, I was working on this as well. Have a look at https://github.com/mmhat/effectful/commit/ed86c24f0e5078fc1f1002374fbf594cba2ad8a9.

Your work seems to be more advanced; I'll have a look at it throughout the next days (probably tomorrow :smile: ).

arybczak:
Nice, your idea is pretty much the same as mine :+1:  The difference that adds complexity in my case is that I account for possibility of collisions of ids returned from `rts_getThreadId` (since that's also what `Hashable ThreadId` uses internally).

I'm not sure if that can happen in practice, but I didn't find a claim that it won't.

mmhat:
> I'm not sure if that can happen in practice, but I didn't find a claim that it won't.

TLDR: No, in practice there should be collissions.

I did some research on this and according to [this](https://gitlab.haskell.org/ghc/ghc/-/issues/16761) discussion it should be Pracitically Impossible. Note that even though the issue is still open the changes relevant to us got merged in https://gitlab.haskell.org/ghc/ghc/-/commit/e57b7cc6d8b1222e0939d19c265b51d2c3c2b4c0.

Code related to this is:

- The `rts_getThreadId` of the RTS: https://github.com/ghc/ghc/blob/0619fb0fb14a98f04aac5f031f6566419fd27495/rts/Threads.c#L180
  Unfortunately this is not exposed; I consider filing an issue in the GHC tracker to change that.
- How a new thread id is obtained in the RTS: https://github.com/ghc/ghc/blob/0619fb0fb14a98f04aac5f031f6566419fd27495/rts/Threads.c#L128
- The counter supplying the RTS with new thread ids: https://github.com/ghc/ghc/blob/0619fb0fb14a98f04aac5f031f6566419fd27495/rts/Threads.c#L32
- The size of that counter: https://github.com/ghc/ghc/blob/0619fb0fb14a98f04aac5f031f6566419fd27495/includes/rts/storage/TSO.h#L25



> `rts_getThreadId` (since that's also what `Hashable ThreadId` uses internally).

I didn't know that there was a `Hashable` instance for `ThreadId`! I came up with my solution by looking at [`GHC.Conc`](https://github.com/ghc/ghc/blob/a7f9670e899bcbc87276446a1aac2304cade2b2f/libraries/base/GHC/Conc/Sync.hs#L155); In particular the `Show` instance of `ThreadId`.

arybczak:
> I did some research on this and according to this discussion it should be Pracitically Impossible.

How so? The program in OP demonstrates the collision. `Show` instance uses `rts_getThreadId`.

The PR https://gitlab.haskell.org/ghc/ghc/-/merge_requests/1254 is a bit weird - `long` is not 64bit long e.g. on x86_32 (which is admittedly going away, but still).

It's also worth noting that the FFI import in both `GHC` and `hashable` is:
```
foreign import ccall unsafe "rts_getThreadId" getThreadId :: ThreadId# -> CInt
```

And `CInt` is a newtype for `Int32`. Looks to me like this wasn't properly fixed.

mmhat:
Well, the OP demonstrated the collision when the thread counter used to be 32bit. Note that https://gitlab.haskell.org/ghc/ghc/-/merge_requests/1254 was never merged into the mainline branch; Instead https://gitlab.haskell.org/ghc/ghc/-/commit/e57b7cc6d8b1222e0939d19c265b51d2c3c2b4c0 was merged. But maybe I didn't get your point, do I?

> And `CInt` is a newtype for `Int32`. Looks to me like this wasn't properly fixed.

This is indeed something that should be fixed.

arybczak:
> Oh, yes, your right! It is not what it said on the tin!

Not to mention that the behavior of the program from OP didn't change at all even with this patch, because FFI import still gets the id truncated to 32 bits, even on x86_64. It's a bit funny that it even works. I'm rusty with C calling conventions, but presumably the result of `rts_getThreadId` ends up in a 64 bit register, but GHC inspects the register that projects the lower 32 bits of said register.

mmhat:
Ah, nice, benchmarks! :+1:

> Anyway, what matters for this PR is that `Hashable` still can produce collisions, so the dance I do in `concUnliftIO` is unfortunately necessary.

This is sad but now I understand at least _why_ you went down this rabbit hole.

FYI: In the meantime I picked up the GHC issue. It seems the issue you described above is scattered all over the RTS.

PR #21, opened 2021-07-09
Added some more instances for Eff

mmhat:
- `Monoid`
- `PrimMonad`
- `Semigroup`

### DIFF
--- a/src/Effectful/Internal/Monad.hs
+++ b/src/Effectful/Internal/Monad.hs
@@ -287,7 +287,7 @@ data PrimE :: Effect where
   PrimE :: PrimE m r
 
 -- | Run an 'Eff' operation with primitive state-transformer actions.
-runPrimE :: Eff (PrimE : es) a -> Eff es a
+runPrimE :: IOE :> es => Eff (PrimE : es) a -> Eff es a
 runPrimE = evalEffect (IdE PrimE)
 
 instance PrimE :> es => PrimMonad (Eff es) where
--- a/src/Effectful/Internal/Monad.hs
+++ b/src/Effectful/Internal/Monad.hs
@@ -95,6 +95,7 @@ type role Eff nominal representational
 -- perform other effects, and it also allows the effects to be handled in any
 -- order.
 newtype Eff (es :: [Effect]) a = Eff (Env es -> IO a)
+  deriving (Monoid, Semigroup)
 
 -- | Run a pure 'Eff' operation.
 --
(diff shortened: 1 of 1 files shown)


mmhat:
I defined `PrimMonad` in terms of the underlying `IO` monad. But we could instead use the `State` effect (or even a separate `PrimEffect`).

arybczak:
`PrimMonad` is fine, but what do the instances for `Semigroup` and `Monoid` do?

arybczak:
I wonder about gating the instance behind the `PrimE` effect and require `IOE` only for `runPrimE`.

In principle it sounds silly since PrimMonad allows you to go *lower* than IO, but on the other hand that would allow to use MutableByteArrayS etc. without full IO access. Well, at least on the surface, but it might still be useful separation as going back to IO requires you to get access to the IO constructor, which would be clearly visible in the code.

mmhat:
I'd also prefer to have a separate effect for that but wanted to have your opinion on this first :+1:

src/Effectful/Internal/Monad.hs
```diff
@@ -272,6 +279,21 @@ instance IOE :> es => MonadBaseControl IO (Eff es) where
...
   withRunInIO = unliftEff
 
+----------------------------------------
+-- Primitive
+
+-- | An effect to perform primitive state-transformer actions
+data PrimE :: Effect where
+  PrimE :: PrimE m r
+
+-- | Run an 'Eff' with primitive state-transformer actions
```
arybczak, line 289:
```suggestion
-- | Run an 'Eff' operation with primitive state-transformer actions.
```

src/Effectful/Internal/Monad.hs
```diff
@@ -272,6 +279,21 @@ instance IOE :> es => MonadBaseControl IO (Eff es) where
 instance IOE :> es => MonadUnliftIO (Eff es) where
   withRunInIO = unliftEff
 
+----------------------------------------
+-- Primitive
+
+-- | An effect to perform primitive state-transformer actions
```
arybczak, line 285:
```suggestion
-- | An effect to perform primitive state-transformer actions.
```

src/Effectful/Internal/Monad.hs
```diff
@@ -272,6 +279,21 @@ instance IOE :> es => MonadBaseControl IO (Eff es) where
...
+data PrimE :: Effect where
+  PrimE :: PrimE m r
+
+-- | Run an 'Eff' with primitive state-transformer actions
+runPrimE :: Eff (PrimE : es) a -> Eff es a
```
arybczak, line 290: This needs `IOE :> es` constraint.

mmhat: Done.

arybczak:
Thanks :)

PR #25, opened 2021-07-13
Added Process effect

mmhat:
This PR adds a `Process` effect used to call run child processes. Note that instead of building upon the [process](https://hackage.haskell.org/package/process) library it uses [typed-process](https://hackage.haskell.org/package/typed-process) in order to provide more type safety.

The idea for this effect came from [this](https://github.com/arybczak/effectful/issues/23#issuecomment-877714896) discussion.

### DIFF
--- a/effectful/effectful.cabal
+++ b/effectful/effectful.cabal
@@ -64,6 +64,7 @@ library
     build-depends:    base >= 4.13 && <5
                     , async
                     , effectful-core >= 0.1 && <0.1.1
+                    , process
                     , resourcet
                     , unliftio
 
@@ -71,6 +72,7 @@ library
 
     exposed-modules: Effectful
                      Effectful.Async
+                     Effectful.Process
                      Effectful.Resource
 
     reexported-modules:    Effectful.Error
--- a/effectful/src/Effectful/Process.hs
+++ b/effectful/src/Effectful/Process.hs
@@ -54,7 +54,7 @@ import qualified System.Process as P
 import Effectful.Internal.Effect
 import Effectful.Internal.Monad
 
--- | An effect for running child processes using the @typed-process@ library.
+-- | An effect for running child processes using the @process@ library.
 data Process :: Effect where
     Process :: Process m r
 
--- a/effectful/src/Effectful/Process.hs
+++ b/effectful/src/Effectful/Process.hs
@@ -1,0 +1,184 @@
+{-# LANGUAGE CPP #-}
+
+module Effectful.Process
+  ( -- * Process effect
+    Process
+  , runProcess
+
+  -- * Running sub-processes
+  , createProcess
+  , createProcess_
+  , P.shell
+  , P.proc
+  , P.CreateProcess(..)
+  , P.CmdSpec(..)
+  , P.StdStream(..)
+  , P.ProcessHandle
+
+  -- ** Simpler functions for common tasks
+  , callProcess
+  , callCommand
+  , spawnProcess
+  , spawnCommand
+  , readCreateProcess
+  , readProcess
+  , readCreateProcessWithExitCode
+  , readProcessWithExitCode
+  , withCreateProcess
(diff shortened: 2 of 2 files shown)


arybczak:
Thanks. I'm not sure if this should go to the main package though :thinking:  Ideally I wanted `effectful` to only depend on boot libraries (+ async since it's very popular anyway). `typed-process` is pretty specific, so this might be good to have as `effectful-typed-process`.

mmhat:
Well, maybe I should have commented a bit more _why_ I opened this PR. Here we go:

I was working under the assumption that we ultimately want `effectful` to include (more or less) the effects mentioned in [this](url) comment. Based on the discussion regarding the `effectful`/`effectful-core` split I totally agree that this effect should life in `effectful-process` and should IMHO be reexported by `effectful` (the batteries-included meta package). Therefore I'd do the following:

1. Wait for the split to happen.
2. Move the Process effect to an own package named `effectful-process`.
3. Reexport it from `effectful`.

Now regarding the questionable choice of `typed-process`: I started the implementation using the `process` package directly (following the decision made in `unliftio`), but I really didn't like the use of `String` for the standard streams. IMHO using a lazy `ByteString` (or at least lazy `Text`) is the correct choice here. I admit that the type variables for the standard streams are optional for this library (and maybe inconvenient at first), but in my experience you deal with those more often than not and at that point those do come in handy.

Note that since `typed-process` itself does only depend on boot libraries the dependency footprint does not grow too much. Whether this is a price we want to pay is a different question.

arybczak:
What I have in mind at the moment wrt. package hierarchy is:

- `effectful-core` - what's currently there.
- `effectful` - `effectful-core` + `async` (since it's ubiquitous) + probably most of the stuff that `unliftio` has, but grouped into effects.

Then `effectful-resourcet` will have the `Resource` effect as `resourcet` integration and it will not be pulled in by `effectful`, because it's pretty specific. Adding dependencies to projects is very cheap, so there is no reason to include a lot of things in `effectful`.

Same with `effectful-typed-process`, it should be a separate library (and can be used interchangeably with the `Process` effect based on the `process` library that'll most likely be included in `effectful`).

I'd like to avoid scope creep wrt. core packages, because they will have to be maintained in the future and my time for open source is limited as is.

It's much better to pass ownership of such packages to other people who have interest in actively maintaining them. E.g. I never used `typed-process`, but you did and are interested enough in using it to have written an `effectful` backend for it, so it should be your package :slightly_smiling_face:

mmhat:
Thanks for clarification! That comment makes it much clearer to me how you envision the scope of the libraries. I feel like that is a good foundation to build on :+1: 

> based on the `process` library that'll most likely be included in `effectful`

I can implement that in the next days; Shouldn't be much different from what I did with `typed-process`.

> my time for open source is limited as is

Feel you.

> other people who have interest in actively maintaining them.

... and I am happy to do so :smiley:
I will contribute that package to the new [effectful-contrib](https://github.com/Kleidukos/effectful-contrib) repository if @Kleidukos is ok with that.

Kleidukos:
@mmhat Hi! I welcome your bindings with open arms in `effectful-contrib`. :)

mmhat:
Here we go :rocket: This one is ready for review.

mmhat:
Regarding `effectful-typed-process`: Unfortunately I won't be able to finish that within the next three weeks. But I guess we are not in a hurry here anyway :slightly_smiling_face:

Kleidukos:
@mmhat I've sent you an invite to be a collaborator on `effectful-contrib`, don't hesitate to open the PR against that repo :)

mmhat:
@Kleidukos Yes, thank you very much! I already accepted the invitation but unfortunately my time is a bit limited in the next weeks :-( Therefore my contribution of `effectful-typed-process` might take a while.

arybczak:
Looks great, thanks :tada:

Issue #20, opened 2021-07-08
Make the initial release


arybczak:
#17 was the last missing piece of the puzzle, all that remains now is some bikeshedding, documentation and a few more tests :slightly_smiling_face: 

Perhaps it would be good to separate `AsyncE` and `Resource` into a different package.

-----

UPDATE (23.01.22):

Left to do:

* [x] Add overview of the library to the `Effectful` module.
* [x] Document modules containing dynamic versions of `Error`, `Reader`, `State` and `Writer`.
* [x] Add tests for `Reader` and `Writer`.
* [x] Move the `Resource` effect to the `resourcet-effectful` package.
* [x] Add `effectful-th` for generation of effect operations as functions for dynamic effects with TH.

mmhat:
I am really looking forward to that! I continued porting one of my RIO-based applications (it's a build system) to this library and I am quite happy with the results: The performance is basically the same, but the code structure improved a lot. So far for the first usage in the wild :smile: 

> Perhaps it would be good to separate `AsyncE` and `Resource` into a different package.

I vote for the `unliftio` approach: `effectful-core`, `effectful-async`, `effectful-resource` and `effectful` being the one with batteries included.

arybczak:
> I am really looking forward to that! I continued porting one of my RIO-based applications (it's a build system) to this library and I am quite happy with the results: The performance is basically the same, but the code structure improved a lot. I didn't find any other bugs though. So far for the first usage in the wild smile

That's great to hear :+1: 

> I vote for the `unliftio` approach: `effectful-core` (and maybe `effectful-async`, `effectful-resource`) and `effectful` being the one with batteries included.

Yes, that sounds like a good approach to me. `effectful-core` for the core and `effectful` for other things. I think I'll mimic `unliftio` and include `async` in `effectful` since it's ubiquitous, but make a separate package for the `Resource` effect.

mmhat:
> but make a separate package for the `Resource` effect.

Do you plan to reexport the `Resource` effect in `effectful`? IMHO it adds some value to the batteries-included version if we were shipping it with some kind of resource effect.

mmhat:
Hm, what are the plans for the `Reader`/`State`/`Writer` effects? Should those live in `effectful` or `effectful-core`? Note that the tests for the `Async` effect require (some of) those. I guess that leaves us with three options:

1. The `Reader`/`State`/`Writer` effects stay in `effectful-core`.
[...]

PR #27, opened 2021-07-17
Split effectful package

mmhat:
As discussed in #20 .


mmhat:
Note that the `Concurrency` and the `Unlift` benchmarks are _not_ (yet) the ones included in the original `effectful` package since those included the `Async` effect which is not part of the package anymore.

arybczak:
Thanks! For simplicity I'd like the tests and benchmarks to be in the `effectful` package only.

You've also duplicated `README.md`, that isn't going to work. You can have a look at the https://github.com/well-typed/optics repo, it has a similar structure (`optics-core`, `optics-extra` and `optics` as an umbrella package) for inspiration.

mmhat:
Thank you for your feedback!

> For simplicity I'd like the tests and benchmarks to be in the `effectful` package only.

Ah, ok, that makes things a lot easier :+1: 

> You've also duplicated `README.md`

I did that because we include the README in the source distribution and I didn't want to change that. However, I did plan to rewrite it, e.g for the `effectful-core` package:
> # effectful-core
> Some short introduction here. Mention and link to the effectful package and how it relates to it.
> For further information have a look at \<link to README.md in the root of the repository>

But I could as well remove the READMEs from the packages; What do you think?

mmhat:
I removed the READMEs from the packages and updated the package descriptions instead. I started a "Usage" section in the README at the root of the repository (Note: I left a _TBD_ in there; I hope that's ok for now).

As a next step I plan to move the `Resource` effect to an own library; But I will do that in a separate PR. Hence it is ready to be reviewed :slightly_smiling_face:

cabal.haskell-ci
```diff
@@ -9,3 +9,6 @@ ghcup-jobs:     >=9.2
 raw-project
   package effectful
     flags: +benchmark-foreign-libraries
+
+  package effectful-core
```
arybczak, line 13: This is not needed anymore.

cabal.project
```diff
@@ -0,0 +1 @@
+packages: */*.cabal
```
arybczak, line 1: Please list cabal files explicitly here.

arybczak (approved):
Looks good modulo 2 minor comments, thanks :+1:

arybczak:
BTW, `README.md` needs update to a link to the filesystem example.

arybczak:
Uh, it looks that anything pushed to the repo will generate a bunch of conflicts now. I'll try not to push more so you can apply the suggestions and fix existing conflicts.

mmhat:
I resolved the issues you mentioned above. Additionally, I added the `FileSystem` example to the test suite: We don't run the code but we will note when we introduced a breaking change and forgot to update it accordingly.

This PR is good to go :rocket:

PR #28, opened 2021-07-26
Add test for nested use of Resource effect

expipiplus1:


effectful/tests/ResourceTests.hs
```diff
@@ -0,0 +1,32 @@
+module ResourceTests
+  ( resourceTests
+  ) where
+
+import           Test.Tasty
```
arybczak, line 5: Please make the import section look like in other files.

expipiplus1:
Without the nesting stuff from the other PR this is probably a little superfluous.

PR #29, opened 2021-07-26
Add useCurrentScope to Resource effect

expipiplus1:
Supersedes https://github.com/arybczak/effectful/pull/28

As first suggested here: https://github.com/snoyberg/conduit/issues/436#issuecomment-616345570

### DIFF
--- a/effectful/effectful.cabal
+++ b/effectful/effectful.cabal
@@ -115,6 +115,7 @@ test-suite test
     other-modules:  AsyncTests
                     ConcurrencyTests
                     ErrorTests
+                    ResourceTests
                     StateTests
                     Utils
                     -- These are examples; We do not actually run them but list
--- a/effectful/src/Effectful/Resource.hs
+++ b/effectful/src/Effectful/Resource.hs
@@ -11,6 +11,9 @@ module Effectful.Resource
   , R.register
   , R.release
   , R.unprotect
+
+  -- * Dealing with nested resource effects
+  , useCurrentScope
   ) where
 
 import Control.Exception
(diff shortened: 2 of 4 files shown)


expipiplus1:
Of course there might be a more general way of accessing "shadowed" effects further up the list, or it might not even make sense at all for this package, I haven't put any thought into it :)

arybczak:
I'm reluctant to accept this based on Michael's comments in the original PR. What's the use case?

expipiplus1:
I've been using that for some time as part of a rendering engine which requires prompt destruction of resources (and not just resources, actions which require begin/end pairs) over many (nested) scopes, it's very handy to be able to do this destruction automatically, while still being able to utilize the global scope.

I don't think that Michael's comments about it being hard to get right are super-valid, after all the (eventual) implementation for `ResourceT` is 2 very simple lines: https://github.com/snoyberg/conduit/issues/436#issuecomment-616345570, and this one is pretty succinct too (I hope I've got the Eff side of it down). 

This can certainly get misused, but I think that's more an argument for clear documentation (happy to add to this PR) rather than total exclusion. 

If this won't make it in, would it at least be possible to expose the `Resource` constructor so that I might publish this in a separate package?

arybczak:
It looks like `getInternalState` and `runInternalState` are enough and if they are in the public API (in the "low level API" section or something), you can write `useCurrentScope` yourself.

Would that be satisfactory?

expipiplus1:
Yeah, that'd be great.

arybczak:
30a28558429bbb20660c17a36d5680afe49f2300 should do the trick :slightly_smiling_face:

expipiplus1:
Thanks!

Issue #23, opened 2021-07-09
Consistent naming of effects


mmhat:
Right now we ship effects with two different naming conventions: Some are suffixed with 'E' (like `AsyncE`) and some are not (like `Reader`). We may consider renaming the effects of one of the two groups in order to have a uniform naming scheme.

mmhat:
Personally I prefer the 'E' suffix for two reasons: The effects are easier to recognize and there are less name collisions with other libraries. (I guess that's why `AsyncE` was named that way in the first place).

arybczak:
There's:
- `IOE` because `IO` is a monad
- `AsyncE` because `Async` represents async actions
- `STME` because `STM` is a monad

There's also a newly added `PrimE`, but arguably it should be `Prim` because it doesn't conflict with anything.

I don't particularly like the `E` suffix and wouldn't want to have `ReaderE`, `StateE` etc :thinking: It's just that for the above three I didn't have a better idea how to name them.

arybczak:
In theory for conflicting types there could be a poly kinded adapter:

```haskell
data family E (a :: k) :: Effect

newtype instance E IO :: Effect where
  IOE :: UnliftStrategy -> E IO m r

data instance E Async :: Effect where
  AsyncE :: E Async m r

data instance E S.STM :: Effect where
  STME :: E S.STM m r

runIO :: Eff '[E IO] a -> IO a
runIO m = unEff (evalEffect (IdE (IOE SeqUnlift)) m) =<< emptyEnv

runAsync :: E IO :> es => Eff (E Async : es) a -> Eff es a
runAsync = evalEffect (IdE AsyncE)

runSTM :: E IO :> es => Eff (E S.STM : es) a -> Eff es a
runSTM = evalEffect (IdE STME)
```

But I'm not sure if it's a good idea.

mmhat:
IMHO the poly kinded adapter does not give us any more consistency compared to the current naming. I have to admit that it took me some time (and having written half a dozen effects) I got used to it and started to value consistency more than the dangling 'E'.

> I don't particularly like the `E` suffix and am not a fan of `ReaderE`, `StateE` etc

Judging from that and [this](https://github.com/arybczak/effectful/pull/22#issuecomment-877633712) comment I guess this is a "won't fix"?

arybczak:
I don't know yet 🤔

arybczak:
Ok, currently we have the following effects:

```
AsyncE
Error
Fail
IOE
Prim
Reader
State
[...]

PR #31, opened 2021-07-29
Add the E suffix to all data types representing effects

arybczak:
Testing the waters for a variant proposed in #23 that advocates adding `E` to the end of all effect data types.

To be fair, I don't really like it, looks messy to me. Thoughts?

### DIFF
--- a/effectful-core/src/Effectful/Error.hs
+++ b/effectful-core/src/Effectful/Error.hs
@@ -20,8 +20,8 @@ For example, if you want to catch an unchecked exception like
 
 -}
 module Effectful.Error
- ( Error
- , runError
+ ( ErrorE
+ , runErrorE
  , throwError
  , catchError
  , tryError
@@ -42,17 +42,17 @@ import Effectful.Internal.Effect
 import Effectful.Internal.Env
 import Effectful.Internal.Monad
 
-newtype Error e :: Effect where
-  Error :: ErrorId -> Error e m r
+newtype ErrorE e :: Effect where
+  ErrorE :: ErrorId -> ErrorE e m r
 
-runError
+runErrorE
   :: forall e es a. Typeable e
-  => Eff (Error e : es) a
+  => Eff (ErrorE e : es) a
   -> Eff es (Either (CallStack, e) a)
-runError m = unsafeEff $ \es0 -> mask $ \release -> do
+runErrorE m = unsafeEff $ \es0 -> mask $ \release -> do
   eid <- newErrorId
   size0 <- sizeEnv es0
-  es <- unsafeConsEnv (IdE (Error @e eid)) noRelinker es0
+  es <- unsafeConsEnv (IdE (ErrorE @e eid)) noRelinker es0
   r <- tryErrorIO release eid es `onException` unsafeTailEnv size0 es
   unsafeTailEnv size0 es
   pure r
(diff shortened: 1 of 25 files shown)


mmhat:
TLDR; I'm ok without the 'E'; We can work around the issue with clever naming and if that's not possible I have no problems with clashing names.

Long answer: As you may have noted I changed my opinion on this topic while discussing #23 :
It since it does only affect `AsyncE` (IOE being the other one; But somehow I don't feel like we should change that) which we might generalize to `Conc` anyway I am fine if we do not an 'E' suffix everywhere.
Note that I chose a different path when I was working on the `typed-process`-based `Process` effect: Instead of using a suffix I went with qualified imports. `polysemy` went with that (e.g. [here](https://hackage.haskell.org/package/polysemy-1.6.0.0/docs/Polysemy-Async.html)) and it seems to work for them. Then the user has to solve the ambiguity problem which IMHO is ok since the needs might be different there. Of course the downside is that we cannot provide a "full" drop-in solution since we cannot reexport all names of the module we are "lifting".
The only thing which annoyed me while implementing the `Process` effect was that `runProcess` was taken by `typed-process`, too.

arybczak:
> polysemy went with that (e.g. here) and it seems to work for them

I don't think `polysemy` is a good template here. For one, the definition of `Async` looks extremely confusing in the docs because `Async` refers to both `Polysemy.Async.Async` and `Control.Concurrent.Async.Async`, but it's hidden unless you hover over links. Also, the effect is sufficiently weird (the `Maybe` in `Async`, no `withAsync` to name a few) that I presume hardly anyone uses it.

> Then the user has to solve the ambiguity problem which IMHO is ok since the needs might be different there

That's not a good approach. Instead of solving the problem once in the library, you just force everyone to solve it individually in their code.

These primitive effects should be a drop-in replacement to the originals, i.e. you import `Effectful.Process.Typed` instead of `System.Process.Typed` and that's it. Otherwise it's just going to be annoying.

arybczak:
> I'm ok without the 'E'; We can work around the issue with clever naming

Yes. I think the rule "Use the suffix `E` only if there are name clashes" is easy to apply and will solve all potential name slashes, because the convention of sticking `E` at the end of type names is not taken by any library.

mmhat:
Yeah, I admit, looking at the docs of `polysemy` again this is indeed confusing... What do you think about extending the `Async` effect to `Conc`? At least for the effects in this repository this would be a "solution by evasion" since we don't have anything left that clashes.

As far as I am concerned that is all what I had in mind when I open #23; I do not expect naming to be consistent across the whole ecosystem.

L-as:
I think it's better to do this universally for consistency.

ocharles:
Personally I think the `E` suffix is the wrong way to go. Only a few effect types have this conflict, but effect types show up frequently in the context of type signatures for things that use events. Given this, I think you want to optimize for effect name readability.

arybczak:
> What do you think about extending the Async effect to Conc? Then we don't have anything left that clashes.

`Conc` collides with the `Conc` data type :smile: `Concurrent` should be fine though.

> Personally I think the E suffix is the wrong way to go. Only a few effect types have this conflict, but effect types show up frequently in the context of type signatures for things that use events. Given this, I think you want to optimize for effect name readability.

Indeed, I think occasional collisions are fine. I'll close this PR for now.

PR #33, opened 2021-09-27
Document static and dynamic effects

mmhat:
It took me a bit to get the difference; Therefore I think a section in the README might be valuable.
The target audience are people who may have never used an effects library before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ readFile path = unsafeEff_ $ do
   -- We are in an IO monad here
   Prelude.readFile path
 
-writeFile :: FileSystem :> es => FilePath -> String -> Eff es ()
+writeFile :: FileSystemStatic :> es => FilePath -> String -> Eff es ()
 writeFile path content = unsafeEff_ $ do
   -- We are in an IO monad here
   Prelude.writeFile path content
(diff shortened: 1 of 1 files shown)


arybczak:
Thanks for writing this.

I'm still unsure about the terminology to use - whether it should be static/dynamic dispatch or primitive/<insert_appropriate_name> effect. Considering that I didn't come up with one of the names for the second option it's most likely better to use static/dynamic, heh.

mmhat:
I like the static/dynamic terminology :slightly_smiling_face:

arybczak:
> I'd appreciate your feedback on the last part where I try to summarize the tradeoffs

Looks good to me :+1: I can't think of any other pros/cons from the top of my head.

Although the "having to deal with internals" part might change. It probably makes sense to move bits necessary for defining a static effect to the public API.

mmhat:
Thank you for your feedback! :+1: 

> Although the "having to deal with internals" part might change. It probably makes sense to move bits necessary for defining a static effect to the public API.

I totally agree with you! Unfortunately I don't have much time right now to work on `effectful` - I am busy writing my Master Thesis, which I plan to finish before February.

arybczak:
Superseded by df9190d3b07da1c788efa1d284f45b61509aebe3.

mmhat:
> Superseded by [df9190d](https://github.com/arybczak/effectful/commit/df9190d3b07da1c788efa1d284f45b61509aebe3).

That's a really nice introduction! :+1:

PR #35, opened 2021-09-27
Add Timeout effect

mmhat:
This PR adds a `Timeout` effect based on [`System.Timeout`](https://hackage.haskell.org/package/base-4.15.0.0/docs/System-Timeout.html).


arybczak:
Thanks :slightly_smiling_face:

PR #36, opened 2021-09-27
Add FileSystem effect

mmhat:
This PR adds a `FileSystem` effect to the `effectful` package as proposed in https://github.com/arybczak/effectful/issues/23#issuecomment-877714896.

### DIFF
--- a/effectful/src/Effectful/FileSystem.hs
+++ b/effectful/src/Effectful/FileSystem.hs
@@ -151,19 +151,15 @@ setCurrentDirectory = unsafeEff_ . D.setCurrentDirectory
 
 -- | Lifted 'D.withCurrentDirectory'.
 withCurrentDirectory :: FileSystem :> es => FilePath -> Eff es a -> Eff es a
-withCurrentDirectory path k = unsafeEff $ \es -> do
-  seqUnliftEff es $ \unlift ->
-    D.withCurrentDirectory path $ unlift k
+withCurrentDirectory path = unsafeLiftMapIO (D.withCurrentDirectory path)
 
 ----------------------------------------
 -- Pre-defined directories
 
--- TODO: Should this have the Environment effect, too?
 -- | Lifted 'D.getHomeDirectory'.
 getHomeDirectory :: FileSystem :> es => Eff es FilePath
 getHomeDirectory = unsafeEff_ D.getHomeDirectory
 
--- TODO: Should this have the Environment effect, too?
 -- | Lifted 'D.getXdgDirectory'.
 getXdgDirectory
   :: FileSystem :> es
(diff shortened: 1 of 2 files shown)


effectful/src/Effectful/FileSystem.hs
```diff
@@ -0,0 +1,357 @@
...
+-- | Lifted 'D.withCurrentDirectory'.
+withCurrentDirectory :: FileSystem :> es => FilePath -> Eff es a -> Eff es a
+withCurrentDirectory path k = unsafeEff $ \es -> do
+  seqUnliftEff es $ \unlift ->
+    D.withCurrentDirectory path $ unlift k
+
+----------------------------------------
+-- Pre-defined directories
+
+-- TODO: Should this have the Environment effect, too?
```
arybczak, line 161: I don't think so :) Same for all TODOs below.

effectful/src/Effectful/FileSystem.hs
```diff
@@ -0,0 +1,357 @@
...
+getCurrentDirectory = unsafeEff_ D.getCurrentDirectory
+
+-- | Lifted 'D.setCurrentDirectory'.
+setCurrentDirectory :: FileSystem :> es => FilePath -> Eff es ()
+setCurrentDirectory = unsafeEff_ . D.setCurrentDirectory
+
+-- | Lifted 'D.withCurrentDirectory'.
+withCurrentDirectory :: FileSystem :> es => FilePath -> Eff es a -> Eff es a
+withCurrentDirectory path k = unsafeEff $ \es -> do
+  seqUnliftEff es $ \unlift ->
```
arybczak, line 155: You can use `unsafeLiftMapIO` here, it'll be a bit simpler. I think it's fine to assume that these functions will not use multiple threads. :thinking: Same below.

effectful/src/Effectful/FileSystem.hs
mmhat: I don't think so, too :)

effectful/src/Effectful/FileSystem.hs
mmhat: Done.

mmhat:
> The (applied) field of effect-tracking systems is still fairly young, especially in the wider Haskell community. I think we also must get comfortable with not getting some things right from the beginning. We will have to learn lessons that time only can teach.

Oh I see! That was the point where I was mistaken: I thought someone out there exists some knowledge about the application of effect systems that I am just not aware of.

Thanks for clarifying; It helped me a lot. :+1: And sorry about the fuss I made by bringing this topic to the table.

arybczak:
Thanks :tada:

PR #38, opened 2021-10-05
Concurrent effect

mmhat:
This PR replaces the `AsyncE` effect with an effect named `Concurrent`.
The code of the former `AsyncE` effect is now located in the `Effectful.Concurrent.Async` module.

In addition to that we add a new module `Effectful.Concurrent.Base` which contains the concurrency primitives found in the `Control.Concurrent` module of the `base` package.

Both `Effectful.Concurrent.Async` and `Effectful.Concurrent.Base` are part of the `Concurrent` effect which lives in an internal module `Effectful.Concurrent.Internal` and gets re-exported by the two modules.

### DIFF
--- a/effectful/bench/Concurrency.hs
+++ b/effectful/bench/Concurrency.hs
@@ -11,7 +11,7 @@ import Test.Tasty.Bench hiding (env)
 
 import Control.Concurrent.Async
 import Control.Monad
-import qualified Effectful.Async as A
+import qualified Effectful.Concurrent.Async as A
 
 import Effectful
 import Utils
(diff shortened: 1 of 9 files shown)


mmhat:
> It seems Environment tests are not well-behaved

That is very weird. Somehow a small mistake made it on master: I accidentally swapped the `expected` and the `actual` arguments of the `assertEqual` calls. I fixed that, but that did not cause the test failure. Regarding those: I checked the code down to the `putenv` C call in `System.Environment` and there is no error wrt. to function calls. Moreover, neither I am able to reproduce it locally nor the CI fails consistently.
Given that we may want to check how the unlifting functions / the tests behave wrt. laziness. Otherwise I have no clue what is going on here.

arybczak:
Thanks!

PR #48, opened 2022-01-05
Tried refining interface

ilyakooo0:
This is meant more as an example to encourage discussion. I fully expect there to be additional iterations on this as a result of discussions.

---

This PR:
1. Solves the unsafety issue of "static" effects by introducing the `EffectStyle` type family.
2. Tried making the interface more encapsulated.

I think the static/dynamic terminology is very clear and understandable. I also think that the "adapter" terminology is confusing and not understandable in the current implementation.

The syntax looks something like this:

```haskell
type instance EffectStyle MyEffect = DynamicEffect

-- or

type instance EffectStyle MyEffect = StaticEffect
```

When defining the effect the user is not actually concerned with how the effect is represented internally (previously `DataA`, `HandlerA`). The user is more concerned with what type of effect he is trying to define. This is precisely what the `EffectStyle` allows the user to specify. In turn, this allows us to determine what is the internal representation of the effect should be. This also allows us to keep the representation consistent between operations of "static" handlers, thus solving the safety problem.

The internal representation of "dynamic" effects is not poking out of the top-level interface at all. The constructor is not even exported.

Unfortunately, the "static" effect constructor still needs to be exported for Haskell type system reasons.

### DIFF
--- a/effectful-core/effectful-core.cabal
+++ b/effectful-core/effectful-core.cabal
@@ -51,6 +51,7 @@ common language
                         TypeApplications
                         TypeFamilies
                         TypeOperators
+                        AllowAmbiguousTypes
 
 library effectful-internal-utils
     import:          language
--- a/effectful-core/src/Effectful.hs
+++ b/effectful-core/src/Effectful.hs
@@ -32,8 +32,10 @@ module Effectful
   , send
 
   -- ** Basic handlers
-  , interpret
-  , reinterpret
+  , runDynamic
+  , rerunDynamic
+  , interpretDynamic
+  , reinterpretDynamic
 
   -- ** Handling local 'Eff' computations
   , LocalEnv
--- a/effectful-core/src/Effectful.hs
+++ b/effectful-core/src/Effectful.hs
@@ -26,16 +26,29 @@ module Effectful
   -- * 'Effect' handlers
   , EffectStyle
   , DynamicEffect
-  , StaticEffect
+  , StaticEffect(..)
 
   -- ** Sending operations to the handler
   , send
 
   -- ** Basic handlers
+  , DynamicEffectHandler
   , runDynamic
   , rerunDynamic
   , interpretDynamic
   , reinterpretDynamic
+
+    -- ** Extending the environment
+  , runStatic
+  , evalStatic
+  , execStatic
+
+    -- ** Static retrieval and update
+  , getStatic
+  , putStatic
+  , stateStatic
+  , stateStaticM
+  , localStatic
 
   -- ** Handling local 'Eff' computations
   , LocalEnv
--- a/effectful-core/src/Effectful.hs
+++ b/effectful-core/src/Effectful.hs
@@ -24,6 +24,9 @@ module Effectful
   , withEffToIO
 
   -- * 'Effect' handlers
+  , EffectStyle
+  , DynamicEffect
+  , StaticEffect
 
   -- ** Sending operations to the handler
   , send
@@ -55,4 +58,5 @@ module Effectful
   ) where
 
 import Effectful.Dispatch.Dynamic
+import Effectful.Dispatch.Static
 import Effectful.Monad
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ dist
 stack.yaml
 stack.yaml.lock
 cabal.project.local
+.vscode
--- a/effectful-core/src/Effectful/Dispatch/Dynamic.hs
+++ b/effectful-core/src/Effectful/Dispatch/Dynamic.hs
@@ -5,7 +5,7 @@ module Effectful.Dispatch.Dynamic
 
   -- * Handling effects
   , EffectHandler
-  , HandlerA(..)
+  , DynamicEffect(..)
   , interpret
   , reinterpret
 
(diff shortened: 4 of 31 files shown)


effectful-core/effectful-core.cabal
```diff
@@ -51,6 +51,7 @@ common language
                         TypeApplications
                         TypeFamilies
                         TypeOperators
+                        AllowAmbiguousTypes
```
arybczak, line 54: This is not a good extension to enable globally.

ilyakooo0:
I also want to note that while this allows us to conceptually simplify the effect style for the user, it is still possible to define custom representations with the same `EffectStyle` type family.

effectful-core/src/Effectful.hs
```diff
@@ -24,13 +24,31 @@ module Effectful
...
+  , DynamicEffect
+  , StaticEffect(..)
 
   -- ** Sending operations to the handler
   , send
 
   -- ** Basic handlers
-  , interpret
-  , reinterpret
+  , DynamicEffectHandler
```
arybczak, line 35: Why did you rename functions for dynamic effects? The suffix `Dynamic` is generally redundant, there is no sensible function that would be named `interpretStatic`.

arybczak (commented):
Thanks for the PR. The main idea is sound (i.e. restricting interface so that you have to use a specific adapter type for each effect).

However, Im not a fan of the fact that dynamic effects need to provide the definition of `EffectStyle` type family. Static effects are supposed to be defined rather seldomly, so it's fine for them as they come with a bunch of caveats anyway, but dynamic are supposed to be "seamless" :thinking: 

EDIT: Maybe it's not that big of a deal, having to write `type instance MyEffect = 'Dynamic` in addition to the definition of an effect is not something particularly tiring.

effectful-core/src/Effectful.hs
ilyakooo0: I think it makes it more obvious what the function is about. 

Yes, there is no corresponding function for static effects, but I don't think having suffix makes it a lot easier for a person to initially grasp the interface. 

---

(I will not fight hard on this, just thought this naming would be more obvious)

effectful-core/effectful-core.cabal
ilyakooo0: Sure, I can remove it all together and add `Proxy` to the place where it is needed.

arybczak: It might be worth experimenting with injective type families, something like

```haskell
data Dispatch = Dynamic | Static

type family DispatchOf (e :: Effect) :: Dispatch

type family AdapterFor (d :: Dispatch) = (r :: Effect -> Type) | r -> d where
  AdapterFor 'Dynamic = HandlerA
  AdapterFor 'Static = DataA

type Adapter e = AdapterFor (DispatchOf e) e

getEnv_
  :: forall e es. e :> es
  => Env es
  -> IO (Adapter e)
getEnv = ...

etc.
```

Then you might not need ambiguous types at all.

effectful-core/src/Effectful.hs
arybczak: Please introduce minimal changes that only introduce safety, no API renaming.

ilyakooo0:
I also have a feeling that having constructors for static effects is in some sense "fake" because the values of the effect type don't actually represent the effect, but rather the data that the effect needs to perform operations. 

Maybe we could start thinking in the direction of making static effects just type declarations without constructors. The type would be used to mark the effect in the monad (just like it does now) and have the underlying data representation defined separately somewhere.

effectful-core/effectful-core.cabal
arybczak: They aren't actually needed, even in this PR as is, must be a remnant from the past.

arybczak:
> Maybe we could start thinking in the direction of making static effects just type declarations without constructors.

I'm not sure what you mean by that.

>  I also think that the "adapter" terminology is confusing and not understandable in the current implementation.

Adapter might not be a good name, it's more of a representation. I'll change it.

ilyakooo0:
> I'm not sure what you mean by that.

> The effect is ultimately what ends up in the type signature, its internal representation doesn't really matter to users.

My point is that in the case of static effects the actual constructor of the effect has nothing to do with the "interface" of the effect. In dynamic effects the interface roughly corresponds to the constructors of the effect data type. In most cases, you can safely export them.

When it comes to static effects, their construct is actually an internal implementation detail, rather than a description of the effect.

It slightly annoys me that the types representing effects have vastly different meanings.

---

Currently, we have this as a definition of a static effect:

```haskell
newtype State s :: Effect where
  State :: s -> State s m r
```

I am proposing something roughly like this:

```haskell
data State s :: Effect

type family StaticEffectRepr (State s) = s
```

The static effect type becomes just a marker in the stack and nothing more. Very clear separation of concerns.

This feels more "genuine" to me for some reason. (At the very least this will be a lot clearer than having two vastly different interpretations of what an effect constructor is.)

---

Before, implementations of dynamic effects were sound with regards to types. The problem was only present in static effects. With the change, I propose above we could restrict the static effects without touching the dynamic effects at all.

---

Another thing that bugged me was that dynamic and static versions of the same effect were not related at all except by name. 

If we make the change I proposed above, instead of the "marker type" of a static effect (the `data` declaration without any constructors) we could use the type of the dynamic version of the effect.

You would still be able to define effects with only a static implementation, effects with only dynamic implementation, and effects with both.

We could introduce some type-level way of distinguishing static vs dynamic version of the effect. For example, static effects could require a `Static` wrapper around it (there is probably a better way of rep[resenting this): 

```haskell
-- | This would be a dynamic effect
foo :: Eff '[State s] ()

-- | And this would be the static effect
bar :: Eff '[Static (State s)] ()
```

---

It seems to me that providing an implementation of a dynamic effect through its' static counterpart is a common pattern.

An added benefit of the approach I described above would be that we could provide a simple interface for exactly this pattern:

```haskell
staticImpl :: (forall x. e x -> Eff '[Static e] x) -> Eff (e : es) a -> Eff es a
```

---

But my primary motivation is clarity of the interface. Having two different `data State s` definitions which represent widely different this is, in my opinion, slightly confusing.

mmhat:
> We could introduce some type-level way of distinguishing static vs dynamic version of the effect.

I am not convinced that this is a good idea: IMHO switching between the dynamic and the static dispatch should be fairly easy; Ideally changing the module imported should be all that needs to be done. I fear that distinguishing the dispatch at the type level makes such a switch impossible without touching a lot of type signatures.
That being said, I do like the `StaticEffectRepr` idea.

arybczak:
Superseded by #49.

PR #49, opened 2022-01-06
Prevent possibility of mixing functions for different dispatch types

arybczak:
Minimal proof of concept based on https://github.com/arybczak/effectful/pull/48#discussion_r779510451.

@ilyakooo0 thoughts?

### DIFF
--- a/effectful-core/src/Effectful/Internal/Monad.hs
+++ b/effectful-core/src/Effectful/Internal/Monad.hs
@@ -54,7 +54,7 @@ module Effectful.Internal.Monad
   , send
 
   -- ** Static dispatch
-  , DataR(..)
+  , DataR
   , runData
   , evalData
   , execData
@@ -157,12 +157,12 @@ unsafeEff_ m = unsafeEff $ \_ -> m
 -- | Get the current 'UnliftStrategy'.
 unliftStrategy :: IOE :> es => Eff es UnliftStrategy
 unliftStrategy = do
-  DataR (IOE unlift) <- getData
+  IOE unlift <- getData
   pure unlift
 
 -- | Locally override the 'UnliftStrategy' with the given value.
 withUnliftStrategy :: IOE :> es => UnliftStrategy -> Eff es a -> Eff es a
-withUnliftStrategy unlift = localData $ \_ -> DataR (IOE unlift)
+withUnliftStrategy unlift = localData $ \_ -> IOE unlift
 
 -- | Create an unlifting function with the current 'UnliftStrategy'.
 --
@@ -273,16 +273,16 @@ instance Fail :> es => MonadFail (Eff es) where
 -- IO
 
 -- | Run arbitrary 'IO' computations via 'MonadIO' or 'MonadUnliftIO'.
-newtype IOE :: Effect where
-  IOE :: UnliftStrategy -> IOE m r
+data IOE :: Effect
 
 type instance DispatchOf IOE = 'Static
+newtype instance DataR IOE = IOE UnliftStrategy
 
 -- | Run an 'Eff' computation with side effects.
 --
 -- For running pure computations see 'runPureEff'.
 runEff :: Eff '[IOE] a -> IO a
-runEff m = unEff (evalData (DataR (IOE SeqUnlift)) m) =<< emptyEnv
+runEff m = unEff (evalData (IOE SeqUnlift) m) =<< emptyEnv
 
 instance IOE :> es => MonadIO (Eff es) where
   liftIO = unsafeEff_
@@ -307,14 +307,14 @@ instance IOE :> es => MonadBaseControl IO (Eff es) where
 -- Primitive
 
 -- | Provide the ability to perform primitive state-transformer actions.
-data Prim :: Effect where
-  Prim :: Prim m r
+data Prim :: Effect
 
 type instance DispatchOf Prim = 'Static
+data instance DataR Prim = Prim
 
 -- | Run an 'Eff' computation with primitive state-transformer actions.
 runPrim :: IOE :> es => Eff (Prim : es) a -> Eff es a
-runPrim = evalData (DataR Prim)
+runPrim = evalData Prim
 
 instance Prim :> es => PrimMonad (Eff es) where
   type PrimState (Eff es) = RealWorld
@@ -330,7 +330,7 @@ data Dispatch = Dynamic | Static
 -- | A dispatch type of a particular effect.
 type family DispatchOf (e :: Effect) :: Dispatch
 
--- | An internal representation of a particular effect.
+-- | Internal representations of effects.
 type family EffectR (d :: Dispatch) = (r :: Effect -> Type) | r -> d where
   EffectR 'Dynamic = HandlerR
   EffectR 'Static  = DataR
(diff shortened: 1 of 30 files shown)


ilyakooo0:
It looks roughly as I imagined it.

The only concern I have is that this PR would restrict the internal representation to exactly two implementations. This is fine in the vast majority of cases, but it is not obvious to me that someone won't encounter a case where there is a dramatically better representation for some corner case.

in #48 the actual representation still remains open.

effectful-core/src/Effectful/Internal/Monad.hs
```diff
@@ -294,14 +310,31 @@ instance IOE :> es => MonadBaseControl IO (Eff es) where
...
 
+----------------------------------------
+-- Dispatch
+
+-- | A dispatch type. For more information consult the documentation in
+-- "Effectful.Dispatch.Dynamic" and "Effectful.Dispatch.Static".
+data Dispatch = Dynamic | Static
+
+-- | A dispatch type of a particular effect.
+type family DispatchOf (e :: Effect) :: Dispatch
```
ilyakooo0, line 331: This is fine as-is, but I have an idea of how we could avoid the need for this altogether: https://github.com/arybczak/effectful/pull/48#issuecomment-1006702437

effectful-core/src/Effectful/Internal/Monad.hs
```diff
@@ -40,15 +41,20 @@ module Effectful.Internal.Monad
 
   -- * Effects
 
+  -- * Dispatch
+  , Dispatch(..)
+  , DispatchOf
+  , Rep
```
ilyakooo0, line 47: I have concerns about exporting this name. It clashes with `Rep` from `GHC.Generic`. If possible, I think we should choose a different (more descriptive) name.

arybczak: This is an internal module, it's not exported from any public one.

effectful-core/src/Effectful/Internal/Monad.hs
arybczak: I don't think you can avoid this, unfortunately. E.g. if you do:

```haskell
newtype DataR :: Effect -> Type where
  DataR :: { unDataR :: StaticRep e } -> DataR e

data family StaticRep (e :: Effect) :: Type
```

and then

```haskell
data State s :: Effect

newtype instance StaticRep (State s) = State s
```

Then you can't use `send`, because there is nothing to send, but you can still mix static and dynamic effects in weird ways.

As for distinguishing static effects from dynamic by having a public wrapper for them, I experimented with it a long time ago and it really clutters type signatures.

arybczak:
> The only concern I have is that this PR would restrict the internal representation to exactly two implementations.

Not quite. Note that types in `Effectful.Internal.Monad` are still unrestricted, so in principle you can introduce a different internal representation and bypass `DispatchOf` completely by using internals.

arybczak:
@ilyakooo0 I turned `DataR` into a data family that has to be defined for any statically dispatched effects, while the effect type itself becomes a dummy. I think it's neat and quite close to what you wanted in https://github.com/arybczak/effectful/pull/48#issuecomment-1006702437.

effectful-core/src/Effectful/Internal/Monad.hs
```diff
@@ -291,17 +307,34 @@ instance IOE :> es => MonadBaseControl IO (Eff es) where
...
+data Dispatch = Dynamic | Static
+
+-- | A dispatch type of a particular effect.
+type family DispatchOf (e :: Effect) :: Dispatch
+
+-- | Internal representations of effects.
+type family EffectR (d :: Dispatch) = (r :: Effect -> Type) | r -> d where
+  EffectR 'Dynamic = HandlerR
+  EffectR 'Static  = DataR
+
```
ilyakooo0, line 337: I really feel like there should be a 

```haskell
type Foo e = EffectR (DispatrchOf e) e
```

ilyakooo0:
I think the change to a data family is really great.

---

I am now of a firm opinion that explicitly declaring that an effect is dynamic is good at least for documentation purposes.

---

I still dislike the naming of `DataR`, `HandlerR`, and `EffectR`. 

I think the meaning of the `R` suffix would not be obvious to an outside observer. I think the clarity of being more descriptive would far outweigh the extra keystrokes.

`Data` seems like a way too generic name and it is totally not obvious how it relates to effects. I think hinting at the fact that it is the internal representation of a static effect would be a good thing.

`HandlerR` – same as above.

I understand that looking from within the library internals the names make sense, but I think the interface of a good library should be oriented towards being easy to understand and encapsulated, rather than catering too heavily to the internals.

**But ultimately, this is not my call to make (:  . I just wanted to state my opinion.**

---

Also, it looks like we are committing to the static/dynamic terminology. I think it is more clear than "local effect", and we might want to rename the modules that implement the effects. (Maybe in a separate PR, but I think this is important)

effectful-core/src/Effectful/Internal/Monad.hs
arybczak: I experimented with that, but I don't think it's worth it. It'd be used in 5 functions, the alias `StaticRepOf` doesn't save that much space and I find type synonyms can be confusing, since GHC expands them somewhat unexpectedly, e.g. in `ghci` `:t` expands type synonyms while `:i` doesn't.

arybczak:
I did some renamings, things should be clearer now.

Overall I'm quite happy with this PR, now it's easier to see and understand the distinction between statically and dynamically dispatched effects from the code and the public API for static dispatch no longer leaks `unsafeCoerce`. Thanks @ilyakooo0 for the ideas and initial proof of concept in #48 :+1: 

> I think it is more clear than "local effect", and we might want to rename the modules that implement the effects

Are you referring to `State` and `Writer` flavors? That's a different thing, dynamic/static is about dispatch, local/shared is about behavior when multiple threads are spawned within the `Eff` monad (see [this](https://github.com/arybczak/effectful/blob/310c1f9bf8982d727301dd5686a710d586f922fd/effectful/src/Effectful/Concurrent/Effect.hs#L9) for more details). In principle the modules could be renamed to `Effectful.State.Static.Local` etc., but I'm not sure about that.

> If there is some way I could contribute, I would be happy to help.

Thanks, though the only thing IMO that needs finishing at this point is remaining documentation, which I have in my head, it's just a matter of finding time and writing it down.

I don't plan any changes to the API at this point, but if you have anything you think is worth improving, let me know.

ilyakooo0:
> Are you referring to State and Writer flavors? That's a different thing, dynamic/static is about dispatch, local/shared is about behavior when multiple threads are spawned within the Eff monad (see this for more details). In principle the modules could be renamed to Effectful.State.Static.Local etc., but I'm not sure about that.


Yeah, I see that. What "shared" meant was quite evident. The confusion arose in my head at the point where I saw `Effectful.State.Dynamic` and by looking at the code it was obvious that it had the implementation of the dynamically dispatched state effect. Right next to it I saw `Effectful.State.Local`. The name made me slightly confused, but looking at the code it became obvious that it was a simple implementation of a statically dispatched state effect.

---

I think that by the design of the library you get the duality of an effect having two different dispatch methods. I think it would be better to make it obvious what is the static counterpart for what dynamic effect.

Seeing three state effects one of which is clearly dynamic and the two others not being obvious creates some confusion IMO.

I think it would be a lot clearer to have:

```
Effectful.State.Dynamic
Effectful.State.Static.Local
Effectful.State.Static.Shared
```

Or alternatively, you could have the "root" module of the effect contain the dynamic effect (the interface), and have the "child" modules be the static (base) implementations of the effect:

```
-- | the dynamic effect (interface of the state effect)
Effectful.State 

-- | the static effects (implementations of the state effect)
Effectful.State.Local
Effectful.State.Shared
```

arybczak:
Let's continue the discussion about module names in #50.

Issue #54, opened 2022-01-24
Port polysemy-plugin and research/improve its usability


arybczak:
It'd be great to port the `polysemy-plugin` to `effectful` to have the option to reduce ambiguity issues if a lot of parametrized effects are in use.

I briefly looked at the source and making it work with effectful should be fairly trivial, but there is a question of its usability. From looking at [polysemy issue tracker](https://github.com/polysemy-research/polysemy/issues?q=plugin+sort%3Acreated-desc) there were multiple tickets about the plugin behaving weirdly, though it looks like recently most (all?) of them were fixed.

Anyone would like to tackle this? I'll not deal with it myself before the initial release.

sproott:
So I've ported the plugin and its tests. Needed to comment out a few tests though because of features that are not in `effectful`. 

The tests all run successfully, so I assume the plugin works as expected. To be honest, I have no idea how the plugin works, I just know how `effectful` works, so I just followed the compiler errors until everything was fine.

[The PR](https://github.com/arybczak/effectful/pull/55)

sproott:
Also, I'm not sure how to handle the license stuff, left the original LICENSE file for now.

arybczak:
Thanks :slightly_smiling_face: 

> To be honest, I have no idea how the plugin works

Yeah, me neither and that's the problem. I'd like to have an understanding of what the plugin does and what it doesn't.

PR #55, opened 2022-02-01
Add effectful-plugin (#54)

sproott:


arybczak:
FWIW the best course of action is to port `cleff-plugin` instead which has much cleaner implementation.

PR #56, opened 2022-02-03
Add effectful-th package

mmhat:
This PR adds a new package `effectful-th` including a TH function `makeSendFunctions` that generates functions for the effect operations of dynamic effects (See the OP of https://github.com/arybczak/effectful/issues/20).

The test suite included just checks if the generated functions for the dynamic effects defined in `effectful-core` compile.

### DIFF
--- a/effectful-th/effectful-th.cabal
+++ b/effectful-th/effectful-th.cabal
@@ -56,8 +56,8 @@ library
     ghc-options:    -O2
 
     build-depends:    base                >= 4.13      && < 5
+                    , containers
                     , effectful-core
-                    , syb
                     , template-haskell
                     , th-abstraction
 
--- a/effectful-th/effectful-th.cabal
+++ b/effectful-th/effectful-th.cabal
@@ -1,0 +1,89 @@
+cabal-version:      2.4
+build-type:         Simple
+name:               effectful-th
+version:            0.1
+license:            BSD-3-Clause
+license-file:       LICENSE
+category:           Control
+maintainer:         andrzej@rybczak.net
+author:             Andrzej Rybczak
+synopsis:           Template Haskell utilities for the effectful library.
+
+description: This package provides several Template Haskell helper functions to
+             aid the development of effects for the 
+             @<https://hackage.haskell.org/package/effectful effectful>@
+             library.
+
+extra-source-files: CHANGELOG.md
+
+tested-with: GHC ==8.8.4 || ==8.10.7 || ==9.0.2 || ==9.2.1
+
+bug-reports:   https://github.com/arybczak/effectful/issues
+source-repository head
+  type:     git
+  location: https://github.com/arybczak/effectful.git
+
+common language
+    ghc-options:        -Wall -Wcompat
+
+    default-language:   Haskell2010
+
+    default-extensions: BangPatterns
+                        ConstraintKinds
+                        DataKinds
+                        DeriveFunctor
+                        DeriveGeneric
+                        FlexibleContexts
+                        FlexibleInstances
+                        GADTs
+                        GeneralizedNewtypeDeriving
+                        LambdaCase
+                        MultiParamTypeClasses
+                        NoStarIsType
+                        RankNTypes
+                        RecordWildCards
+                        RoleAnnotations
+                        ScopedTypeVariables
+                        StandaloneDeriving
+                        TupleSections
+                        TypeApplications
+                        TypeFamilies
+                        TypeOperators
+
+library
+    import:         language
+
+    ghc-options:    -O2
+
+    build-depends:    base                >= 4.13      && < 5
+                    , effectful-core
+                    , syb
+                    , template-haskell
+                    , th-abstraction
+
+    hs-source-dirs:  src
+
+    exposed-modules: Effectful.TH
+
+test-suite effectful-th-test
+    import:         language
+
+    ghc-options:    -threaded -rtsopts -with-rtsopts=-N4
+
+    build-depends:    base
+                    , effectful-core
+                    , effectful-th
+
+    hs-source-dirs: tests examples
+
+    type:           exitcode-stdio-1.0
+    main-is:        Main.hs
+
+    other-modules:  Error
+                    Fail
+                    Reader
+                    State
+                    Writer
+                    -- These are examples; We do not actually run them but list
+                    -- them here in order to check if they compile.
+                    MyReader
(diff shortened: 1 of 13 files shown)


mmhat:
@arybczak I think this one can be reviewed and merged. :slightly_smiling_face:

effectful-th/src/Effectful/TH.hs
```diff
@@ -0,0 +1,238 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE TemplateHaskell #-}
```
arybczak, line 2: Can you replace this with [TemplateHaskellQuotes](https://mpickering.github.io/ghc-docs/build-html/users_guide/glasgow_exts.html#extension-TemplateHaskellQuotes)?

effectful-th/src/Effectful/TH.hs
```diff
@@ -0,0 +1,238 @@
...
+-- | Template Haskell utilities.
+module Effectful.TH
+  ( -- * Generate functions for dynamic effects
+    makeSendFunctions
```
arybczak, line 6: Naming is hard, I'm wondering whether it should be `makeEffect` as in other libraries. The problem is of course that it's only for dynamic dispatch.

effectful-th/src/Effectful/TH.hs
```diff
@@ -0,0 +1,238 @@
...
+  , makeSendFunctionsWithOptions
+  , makeSendFunctionFor
+  , Options
+  , defaultOptions
+  , setMakeFunction
+  , setMakeFunctionFor
+  , setToFunctionName
+
+    -- * Re-exports
```
arybczak, line 15: Why the re-exports?

effectful-th/src/Effectful/TH.hs
```diff
@@ -0,0 +1,238 @@
...
+
+-- | Generate functions for sending an effect operation to its handler.
+--
+-- For example,
+--
+-- > -- These extensions are needed for the definition of the effect.
+-- > {-# LANGUAGE DataKinds #-}
+-- > {-# LANGUAGE GADTs #-}
+-- > {-# LANGUAGE TypeFamilies #-}
+-- > -- These extensions are needed for the generation of the functions.
```
arybczak, line 80: You can check in TH whether these are enabled and complain otherwise, see https://github.com/well-typed/optics/blob/82f4b8c4957e927147c7ca61b86d9629d1894ea5/optics-th/src/Optics/TH/Internal/Utils.hs#L112 and how it's used.

effectful-th/effectful-th.cabal
```diff
@@ -0,0 +1,89 @@
...
+                        StandaloneDeriving
+                        TupleSections
+                        TypeApplications
+                        TypeFamilies
+                        TypeOperators
+
+library
+    import:         language
+
+    ghc-options:    -O2
```
arybczak, line 56: This is not needed :)

effectful-th/effectful-th.cabal
```diff
@@ -0,0 +1,89 @@
...
+
+    build-depends:    base                >= 4.13      && < 5
+                    , containers
+                    , effectful-core
+                    , template-haskell
```
arybczak, line 61: Need to provide upper and lower version bound, it's very likely something breaks in GHC 9.4.

arybczak (commented):
Thanks!

Do other packages have more tests for a bit more "weird" effect data types that can be added?

For example, I wonder if this works: https://github.com/polysemy-research/polysemy/issues/339

mmhat:
Unfortunately the current implementation does not work well with "weird" effects.
This module is build around `th-abstraction` - which is just too much abstraction to support these types.
In particular, both
```
data T1 a where
  C1 :: forall a b. T1 a

data T2 a where
  C2 :: forall b a. T2 a
```
are reified as to something like:
```
DatatypeInfo { ..., datatypeVars = [KindedTV a () StarT], ..., datatypeCons = [ConstructorInfo {..., constructorVars = [KindedTV b () StarT], ...}]}
```
Note the order of the type variables of the explicit forall is completely lost and we cannot figure out how to construct the type applications at the call site.

I will think about a more elaborate solution using `template-haskell` directly.

arybczak:
I wouldn't worry about that, it's better to use `th-abstraction` as that eases maintenance burden, direct `template-haskell` breaks too often.

mmhat:
> I wouldn't worry about that, it's better to use `th-abstraction` as that eases maintenance burden, direct `template-haskell` breaks too often.

That's why I chose it in the first place :slightly_smiling_face: 
However, the example given in https://github.com/polysemy-research/polysemy/issues/339 will not work since it requires an extra type application of the `ty` variable and I don't know (yet) how to get this right.

I propose the following: Unless I find a clear path forward throughout the next days will push some more documentation regarding known limitations and we can merge this (I will of course address the remaining issues first). If we figure out a way to improve the module in this regard it can happen on a separate PR.

effectful-th/src/Effectful/TH.hs
mmhat: Unfortunately that is not possible. Changing that to `TemplateHaskellQuotes` results in:
```
src/Effectful/TH.hs:156:40: error:
    Operator applied to too few arguments: :>
    |
156 |       effectConstraint = [t| $(effect) :> $(varT es) |]
    |
```

effectful-th/src/Effectful/TH.hs
mmhat: Yeah, I am not really happy with the naming. `makeEffect` is nice; I will change that.

effectful-th/src/Effectful/TH.hs
mmhat: I included them for slightly better ergonomics: This way one can just import `Effectful.TH` and is not confronted with a bunch of "... not in scope" error messages immediately.

effectful-th/src/Effectful/TH.hs
mmhat: Oh, nice! I'll have a look.

effectful-th/effectful-th.cabal
mmhat: Whoops... removed.

effectful-th/src/Effectful/TH.hs
arybczak: This requires getting rid of splices, but it should be possible as you can construct the AST manually. FWIW `optics-th` uses TemplateHaskellQuotes only. This matter for cross-compilation.

arybczak:
I think I'll go with #60. The code is trivially ported from `cleff`.

arybczak:
Superseded by #60, thanks for the work though!

Issue #57, opened 2022-02-08
How to use other effects in `Static` effects?


oberblastmeister:
The documentation explains that you can run arbitrary `IO` actions using `unsafeEff_` as long as the effects are reintroduced in function that runs the effect. How can you run `effectful` effects such as `Error`, `Writer`, etc. in the same way? Is it safe to use `unsafeCoerce` to coerce the type level effects as long as they are introduced in the run function?

arybczak:
> Is it safe to use unsafeCoerce to coerce the type level effects as long as they are introduced in the run function?

No.

Static effects are primitive, all they are supposed to be able to do is manipulation of the StaticRep and IO. If you want to reuse existing effects, you should use dynamic dispatch and `reinterpret`.

oberblastmeister:
Could we mitigate the usage of `unsafeEff` by having a type family for static effects that says if it requires `IO` or not. Then in functions such as `evalStaticRep` we could force the introduction of `IOE`. For example something like this

```haskell
data FileSystem :: Effect

type NeedsIO :: Effect -> Bool
type family NeedsIO e

type instance NeedsIO FileSystem = 'True

deferIO :: (DispatchOf e ~ 'Static, NeedsIO e ~ 'True) => IO a -> Eff es a
deferIO = unsafeEff_

evalStaticRepDeferred :: (DispatchOf e ~ 'Static, NeedsIO e ~ 'True, IOE :> es) => StaticRep e -> Eff (e ': es) a -> Eff es a
evalStaticRepDeferred = evalStaticRep
```

arybczak:
Hmm. Yeah, that seems like a good idea :+1:

arybczak:
Fixed in b8c3711cbe6246bb8d583b2796d3d937d2bd9eaa.

PR #58, opened 2022-02-08
make IO hidden in static effect checked at compile time

oberblastmeister:
Addresses #57 and Kleidukos/effectful-contrib#23
Makes it so that `unsafeEff_` is not needed anymore to run `IO` actions inside of a static effect. The `IOE` effect will automatically be introduced in the running function if `IO` needs to be used inside the effect

### DIFF
--- a/effectful-core/effectful-core.cabal
+++ b/effectful-core/effectful-core.cabal
@@ -51,6 +51,7 @@ common language
                         TypeApplications
                         TypeFamilies
                         TypeOperators
+                        StandaloneKindSignatures
 
 library effectful-internal-utils
     import:          language
--- a/effectful-core/src/Effectful/Dispatch/Static.hs
+++ b/effectful-core/src/Effectful/Dispatch/Static.hs
@@ -56,6 +56,8 @@ module Effectful.Dispatch.Static
   , putEnv
   , stateEnv
   , modifyEnv
+  , NeedsIO
+  , deferIO
   ) where
 
 import Effectful.Internal.Env
--- a/effectful-core/src/Effectful/Dispatch/Static.hs
+++ b/effectful-core/src/Effectful/Dispatch/Static.hs
@@ -58,6 +58,7 @@ module Effectful.Dispatch.Static
   , modifyEnv
   , NeedsIO
   , deferIO
+  , deferIO_
   ) where
 
 import Effectful.Internal.Env
--- a/effectful-core/src/Effectful/Internal/Monad.hs
+++ b/effectful-core/src/Effectful/Internal/Monad.hs
@@ -66,12 +66,16 @@ module Effectful.Internal.Monad
 
   -- *** Primitive operations
   , unsafeConsEnv
+  , unsafeRunStaticRep
+  , unsafeEvalStaticRep
+  , unsafeExecStaticRep
   , getEnv
   , putEnv
   , stateEnv
   , modifyEnv
   , NeedsIO
   , deferIO
+  , deferIO_
   ) where
 
 import Control.Applicative (liftA2)
@@ -321,7 +325,7 @@ runPrim = evalStaticRep Prim
 
 instance Prim :> es => PrimMonad (Eff es) where
   type PrimState (Eff es) = RealWorld
-  primitive = deferIO (Proxy @Prim) . IO
+  primitive = deferIO_ (Proxy @Prim) . IO
 
 ----------------------------------------
 -- Dispatch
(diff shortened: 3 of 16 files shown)


oberblastmeister:
@arybczak Do we need `deferIO` with the `Env`?. Also should I use `AllowAmbiguousTypes` instead of `Proxy`?

arybczak:
Eh, the ambiguous type / the need to use proxy is a serious issue :disappointed:

arybczak:
> I think most effectful functions are ambiguous anyway

?

oberblastmeister:
I mean stuff like `get` and `put`. They don't need ambiguous types, but mostly you have to instantiate the types unless you use some sort of plugin I think.

arybczak:
> but mostly you have to instantiate the types

Not really, because `get` returns a state and `put` takes it as an argument, so if these types are known (and they are most of the time), then you don't need annotations.

This is different since you always have to specify the effect.

I'm wondering if a compromise would work - augment `Static` with a parameter `data Purity = Pure | Impure` and require `IOE` for `Impure` ones, but using `IO` would still be via `unsafeEff_`.

oberblastmeister:
> so if these types are known (and they are most of the time), then you don't need annotations.

Oh yeah, you are right. I guess it wasn't working for me because I was using `optics` with overloaded labels.

> I'm wondering if a compromise would work - augment Static with a parameter data Purity = Pure | Impure and require IOE for Impure ones, but using IO would still be via unsafeEff_.

I guess that would be a little better, but still not fully safe.

oberblastmeister:
IMO it would be better to have more safety than having less typing, you did say static effects are primitive

arybczak:
Another problem here is that now any function that pulls in a static effect marked `NeedsIO` can call `deferIO`. This is not what we want, `IO` operations should be constrained to `core` operations of the effect.

Btw, why did you name this function `deferIO`?

oberblastmeister:
> Btw, why did you name this function deferIO? It's not deferring anything.

I named it because it "defers" the `IOE` constraint until you run the effect. I didn't really think of the names when writing the code though, we could probably pick something better

oberblastmeister:
I wonder if it is a good idea in general to allow one to hide `IO` until the running function. If you want to wrap a logger that uses `IO`, I think it makes sense because you would not want `IO` to infect your pure code but you might still want to log. However, for effects like `FileSystem`, `Temporary`, etc. I am probably doing `IO` anyway. So having separate effects just makes everything more verbose. I have longer signatures and I have to call more functions to run the effects. Furthermore, if I want to use a function that accesses the filesystem that has not been wrapped by Effectful, now I have more effects, both `IOE` and `FileSystem`. Why not just require the effects to have `IOE`? Using that logic, I wonder if it would be a good idea to just let hiding `IO` be unsafe. This would discourage people to hide it. We could still make it a little safer by augmenting `Static` with a parameter. What are your thoughts on this?

arybczak:
> However, for effects like FileSystem, Temporary, etc. I am probably doing IO anyway.

Yes, but ideally the IO you're doing is encapsulated within other effects.

> So having separate effects just makes everything more verbose. I have longer signatures and I have to call more functions to run the effects. 

It's a compromise. You get longer signatures, but you can see from the signature more things about what the function does and what it doesn't. This matters in any application that is being worked on by multiple people, has >100k LoC and some legacy baggage.

> Furthermore, if I want to use a function that accesses the filesystem that has not been wrapped by Effectful, now I have more effects, both IOE and FileSystem.

You can do two things:
1) Have your application locally extend the `FileSystem` effect with the relevant wrapper using `unsafeEff` while you submit a PR to effectful.
2) Ignore these granular wrappers and use `UnliftIO` directly.

> I wonder if it is a good idea in general to allow one to hide IO until the running function. [...] Using that logic, I wonder if it would be a good idea to just let hiding IO be unsafe. 

Do you think it's the same for dynamically dispatched effects? Because it's not and it's the same idea as for static dispatch, just the implementation is different. It's one of the main perks of the effect systems - to not expose the full power of IO to everything but interpreters (for dynamic dispatch) and core operations (static dispatch). It's just less principled for static dispatch.

oberblastmeister:
Ok, that makes sense. Thanks for explaining to me.

arybczak:
I added b8c3711cbe6246bb8d583b2796d3d937d2bd9eaa - I think this is a good compromise. I thought about skipping it entirely, but then I wrote a proof of concept and found out that I missed the `IOE` constraint in `runInternalState` :slightly_smiling_face:

oberblastmeister:
Nice, I think it is a good compromise.

PR #60, opened 2022-02-19
Add effectful-th

arybczak:
Supersedes #56, based on the code from `cleff` and #56.

Doesn't need `TemplateHaskell`, works with more data types and produces arguably prettier type signatures.

### DIFF
--- a/effectful-th/src/Effectful/TH.hs
+++ b/effectful-th/src/Effectful/TH.hs
@@ -1,0 +1,265 @@
+{-# LANGUAGE CPP #-}
+{-# LANGUAGE TemplateHaskellQuotes #-}
+-- | Generate functions for performing operations of dynamically dispatched
+-- effects via Template Haskell.
+module Effectful.TH
+  ( makeEffect
+  , makeEffect_
+  ) where
+
+import Control.Monad
+import Data.Char (toLower)
+import Data.Foldable (foldl')
+import Data.Maybe
+import Language.Haskell.TH
+import Language.Haskell.TH.Datatype
+import Language.Haskell.TH.Datatype.TyVarBndr
+import qualified Data.Map.Strict as Map
+
+import Effectful
+import Effectful.Dispatch.Dynamic
+
+-- | For an effect data type @E@, @'makeEffect' E@ generates the appropriate
+-- instance of 'DispatchOf' as well as functions for performing operations of
+-- @E@ by 'send'ing them to the effect handler.
+--
+-- >>> :{
+--   data E :: Effect where
+--     Op1 :: Int -> m a -> E m a
+--     Op2 :: IOE :> es => Int -> E (Eff es) ()
+--     Op3 :: (forall r. m r -> m r) -> E m Int
+--   makeEffect ''E
+-- :}
+--
+-- >>> :kind! DispatchOf E
+-- DispatchOf E :: Dispatch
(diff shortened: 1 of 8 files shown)


arybczak:
@mmhat I've taken the base from `cleff`, but it didn't support a few cases I would've liked, so I modified it (also applied a few things from your PR) and now it generates very good type signatures and works for truly bizarre effect data types.

effectful-th/src/Effectful/TH.hs
```diff
@@ -0,0 +1,243 @@
...
+
+toSmartConName :: String -> String
+toSmartConName = \case
+  (':' : xs) -> xs
+  (x : xs)   -> toLower x : xs
+  _          -> error "empty constructor name"
+
+extractCtx :: Type -> ([TyVarBndrSpec], Cxt)
+extractCtx = \case
+  ForallT vars ctx _ -> (vars, ctx)
```
re-xyr, line 209: Is this not recursive because the reified type will definitely be normalized?

effectful-th/src/Effectful/TH.hs
```diff
@@ -0,0 +1,243 @@
...
+  -- e :: E v1 .. vN :> es => x1 -> .. -> xK -> E v1 .. vN (Eff es) r
+  -- e x1 .. xK = send (E x1 .. xN :: E v1 .. vK (Eff es) r)
+  let fnBody =
+        let effOp  = foldl' (\f -> AppE f . VarE) (ConE name) fnArgs
+            effSig = effTy `AppT` (ConT ''Eff `AppT` esVar) `AppT` substM resTy
+        in if makeSig
+           then VarE 'send `AppE` SigE effOp effSig
+           else VarE 'send `AppE`      effOp
+#endif
+  let fnSig = ForallT actionVars
```
re-xyr, line 190: If we use explicit quantification, datatype heads without kind annotations for their parameters like `data Effect m a` will fail with confusing type ambiguity errors under `PolyKinds`. One way to solve this is to scrap explicit tyvars and let the inference do the job, at the cost of not being able to generate some ambiguous constructors. What do you think?

effectful-th/src/Effectful/TH.hs
arybczak: GADT constructors do not allow nested foralls: https://gitlab.haskell.org/ghc/ghc/-/issues/18191

effectful-th/src/Effectful/TH.hs
arybczak: They will fail, but there is a check earlier in `makeEffectImpl` that validates kinds of the last two type parameters, so if you skip it, you will get a reasonable error message:

```
tests/ThTests.hs:21:1: error:
    Expected the next to last type parameter to have a kind * -> *, got k_0
   |
21 | makeEffect ''SimpleADT
```

```
tests/ThTests.hs:21:1: error:
    Expected the last type parameter to have a kind *, got k_0
   |
21 | makeEffect ''SimpleADT
   | ^^^^^^^^^^^^^^^^^^^^^^
```

This encourages writing effects like this:

```
data E x1 ... xN :: Effect where
   ...
```

which I find more informative :slightly_smiling_face:

mmhat:
Nice work! :+1: 
This is definitely better than my efforts in #56 .

Issue #69, opened 2022-07-18
How to interoperate with concurrent monad transformers


sullyj3:
I'm trying to use `effectful` together with `streamly`. I have a `SerialT (Eff es) a`, which uses streamly's concurrency combinators internally ([`async`](https://hackage.haskell.org/package/streamly-0.8.2/docs/Streamly-Prelude.html#v:async) and [`fromAsync`](https://hackage.haskell.org/package/streamly-0.8.2/docs/Streamly-Prelude.html#v:fromAsync)). I'm getting the runtime error:

```
If you want to use the unlifting function to run Eff computations in multiple threads, have a look at UnliftStrategy (ConcUnlift).
```
I had a look at the docs for `ConcUnlift`, but I'm struggling to understand how and where to use it. I had hoped I could just use it at the the entry point to my program, and maybe it would work but just be a little slower than if it was targeted more specifically:
```
-- (simplified)
main :: IO ()
main = do
  config <- getConfig
  runEff .
    withUnliftStrategy (ConcUnlift Ephemeral Unlimited) .
    runFailIO .
    runReader config $
    myProgram
```

But it doesn't seem to solve the problem - it still results in the same runtime error. Would it be possible to get some relevant examples in the documentation?

arybczak:
Can you check if https://github.com/haskell-effectful/effectful/pull/70 fixes the problem?

arybczak:
I'm pretty sure not resetting the strategy is the correct choice either way, because if it's reset, then it's impossible to re-use external pieces of code that use unlifting and spawn more than one level of threads.

For some reason I didn't think of this scenario before :disappointed:

sullyj3:
Works perfectly, thank you! Although I don't really understand why, since I'm not calling `withEffToIO`. Is that happening under the hood somewhere?

arybczak:
> Is that happening under the hood somewhere? 

Yes, instances of `MonadUnliftIO` / `MonadBaseControl IO` use this function.

sullyj3:
Ah, that makes sense.

PR #70, opened 2022-07-18
Don't reset the strategy inside the continuation in withEffToIO

arybczak:
Fixes https://github.com/haskell-effectful/effectful/issues/69.


Kleidukos:
Perhaps a changelog entry?

arybczak:
I planned to do this later, but sure, why not. I also bumped the version.

PR #89, opened 2022-08-18
Improve callstacks when using the dynamic error effect

EggBaconAndSpam:
When using `throwError` from `Effectful.Error.Dynamic` the call site of interest is buried under 4 lines of internal implementation details: 

```
> Left (cs,_) <- runEff $ runError @() $ throwError ()
> putStrLn $ prettyCallStack cs
CallStack (from HasCallStack):
  throwError, called at src/Effectful/Error/Dynamic.hs:43:21 in effectful-core-2.1.0.0-1rwlVbEDo1s77mEaNtxUZU:Effectful.Error.Dynamic
  handler, called at src/Effectful/Dispatch/Dynamic.hs:371:50 in effectful-core-2.1.0.0-1rwlVbEDo1s77mEaNtxUZU:Effectful.Dispatch.Dynamic
  handle, called at src/Effectful/Internal/Monad.hs:410:10 in effectful-core-2.1.0.0-1rwlVbEDo1s77mEaNtxUZU:Effectful.Internal.Monad
  send, called at src/Effectful/Error/Dynamic.hs:60:14 in effectful-core-2.1.0.0-1rwlVbEDo1s77mEaNtxUZU:Effectful.Error.Dynamic
  throwError, called at <interactive>:1:48 in interactive:Ghci1
```

We can fix this by making the `CallStack` an explicit event argument, then:
```
> Left (cs,_) <- runEff $ runError @() $ throwError ()
> putStrLn $ prettyCallStack cs
CallStack (from HasCallStack):
  throwError, called at <interactive>:12:48 in interactive:Ghci7
```
This matches the behaviour of `Effectful.Error.Static`.

### DIFF
--- a/effectful-core/src/Effectful/Error/Dynamic.hs
+++ b/effectful-core/src/Effectful/Error/Dynamic.hs
@@ -27,10 +27,11 @@ module Effectful.Error.Dynamic
 import Effectful
 import Effectful.Dispatch.Dynamic
 import qualified Effectful.Error.Static as E
+import GHC.Stack (callStack)
 
 -- | Provide the ability to handle errors of type @e@.
 data Error e :: Effect where
-  ThrowError :: e -> Error e m a
+  ThrowError :: E.CallStack -> e -> Error e m a
   CatchError :: m a -> (E.CallStack -> e -> m a) -> Error e m a
 
 type instance DispatchOf (Error e) = Dynamic
@@ -40,7 +41,7 @@ runError
   :: Eff (Error e : es) a
   -> Eff es (Either (E.CallStack, e) a)
 runError = reinterpret E.runError $ \env -> \case
-  ThrowError e   -> E.throwError e
+  ThrowError theCallStack e   -> E.throwError' theCallStack e
   CatchError m h -> localSeqUnlift env $ \unlift -> do
     E.catchError (unlift m) (\cs -> unlift . h cs)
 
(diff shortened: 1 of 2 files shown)


arybczak:
Hmm. Why not

```haskell
diff --git a/effectful-core/src/Effectful/Error/Dynamic.hs b/effectful-core/src/Effectful/Error/Dynamic.hs
index 59080de..782f098 100644
--- a/effectful-core/src/Effectful/Error/Dynamic.hs
+++ b/effectful-core/src/Effectful/Error/Dynamic.hs
@@ -24,6 +24,8 @@ module Effectful.Error.Dynamic
   , E.prettyCallStack
   ) where
 
+import GHC.Stack (withFrozenCallStack)
+
 import Effectful
 import Effectful.Dispatch.Dynamic
 import qualified Effectful.Error.Static as E
@@ -57,7 +59,7 @@ throwError
   => e
   -- ^ The error.
   -> Eff es a
-throwError = send . ThrowError
+throwError e = withFrozenCallStack (send $ ThrowError e)
 
 -- | Handle an error of type @e@.
 catchError
```

instead?

Also, I wonder if this makes sense in general. Maybe there are situations in which it's good to know what exactly throws in the current implementation of `ThrowError`?

EggBaconAndSpam:
Oh, I didn't know about frozen call stacks -- that's a much simpler solution! I'm away from my PC today so I'll close this for now. If you don't I'll create a PR with your proposed diff next week.

As to why this is useful at all, I'm not sure why anyone would use the dynamic error effect in the first place but I do believe that they wouldn't expect to see 4 lines of internals in their stack traces.

arybczak:
Fair enough. I've added this in c99d1bdfae12dc3c3fee43aeea3ad49268fbbeac.

PR #94, opened 2022-08-28
Disallow arbitrary IO in Prim

qrpnxz:
By using the `RealWorld` state thread, `Prim` allows arbitrary `IO` execution:

```hs
ioToEff :: Prim :> es => IO a -> Eff es a
ioToEff = Control.Monad.Primitive.ioToPrim
```

Fixed by using it's own state thread.


arybczak:
I'm not sure if this is safe. In theory it is since State# doesn't exist at runtime, but :thinking:

qrpnxz:
According to `GHC.Exts`, `State#` is a `ZeroBitType` and "the only purpose of the type parameter is to keep different state threads separate. It is represented by nothing at all." Given that, it should be safe.

arybczak:
Fair enough, this can always be reverted if it blows up in any way. I created a bit more elaborate version [here](#95).

Issue #93, opened 2022-08-24
Cannot run `Prim` effect without invoking `IOE`


qrpnxz:
`Prim` currently has me either announcing I used `Prim`, or worse, if I try to get rid of `Prim`, say I'm doing arbitrary `IO`. I would like to be able to use `Prim` without requiring the effect from callers.

Would this be possible?:

```hs
runPrim :: (forall s. Eff [Prim s : es] a) -> Eff es a
runPrimIO :: IOE :> es => Eff [Prim RealWorld : es] a -> Eff es a
```

arybczak:
Unfortunately not. This doesn't work:

```haskell
data Prim s :: Effect

type instance DispatchOf (Prim s) = Static WithSideEffects
data instance StaticRep (Prim s) = Prim

instance Prim s :> es => PrimMonad (Eff es) where
  type PrimState (Eff es) = s
  primitive = unsafeEff_ . IO . unsafeCoerce
```

because:

```
error:
    The RHS of an associated type declaration mentions out-of-scope variable ‘s’
      All such variables must be bound on the LHS
    |
310 |   type PrimState (Eff es) = s
```

To have Prim be `ST`-like `Eff` would have to be given additional type parameter. I experimented with this in the past, but decided against this as it has limited use, yet imposes a cost for everyone.

qrpnxz:
I came up with something similar:

```hs
data STE s :: Effect

type instance DispatchOf (STE s) = Static NoSideEffects
data instance StaticRep (STE s) = STE

runSTE :: (forall s. Eff (STE s : es) a) -> Eff es a
runSTE eff = evalStaticRep STE eff

instance PrimMonad (Eff (STE s : es)) where
  type PrimState (Eff (STE s : es)) = s
  primitive = unsafeEff_ . IO . unsafeCoerce
```

I just haven't tested ergonomics yet in real code because overlapping with the current instance makes it a pain, but this should be pretty decent I think.

qrpnxz:
I had an idea of a type family that we could put on the RHS that would match on any place on the layer stack, but have not tried it yet.

arybczak:
> I just haven't tested ergonomics yet in real code because overlapping with the current instance makes it a pain, but this should be pretty decent I think.

There are two issues here:

1. It's not possible to use this with a typical/encouraged `STE s :> es => Eff es ()` pattern.
[...]

Issue #96, opened 2022-08-31
Effect idea: Lifetime tracking with effects


qrpnxz:
Building on the ideas from #93's `STE`, turns out we can use `Effects` to track object lifetimes:

```hs

data Life s :: Effect
type instance DispatchOf (Life s) = Static NoSideEffects
data instance StaticRep (Life s) = Life

newtype Tagged s a = Tagged a

with :: (forall s. Tagged s r -> Eff (Life s : es) a) -> r -> Eff es a
with k = evalStaticRep Life . k . Tagged

withFile :: IOE :> es => FilePath -> IOMode -> (forall s. Tagged s Handle -> Eff (Life s : es) a) -> Eff es a
withFile path mode k = bracket (openFile path mode) hClose (with k)

hFileSize :: (IOE :> es, Life s :> es) => Tagged s Handle -> Eff es Integer
hFileSize (Tagged handle) = liftIO $ hFileSize handle

hTell :: (IOE :> es, Life s :> es) => Tagged s Handle -> Eff es Integer
hTell (Tagged handle) = liftIO $ hTell handle
```

Example:
```hs
main = runEff $
  withFile "/usr/bin/bash" ReadMode $ \bashHandle ->
    withFile "/usr/bin/grep" ReadMode $ \grepHandle -> do
      (liftIO . print) =<< hFileSize grepHandle
      (liftIO . print) =<< raise (hFileSize bashHandle)
```
(`raise` to avoid overlapping instances. Perhaps the plug-in helps here, but I haven't been able to use it.)

---

This is utterly inescapable as far as I can tell. You can try to sneak the value out with a `GADT`, but you can't do anything with it because the effect is gone.

Kleidukos:
Hi! I am not too sure about the specifics, but I can see that it could be put to good use with the FS and Process effects :)

arybczak:
> This is utterly inescapable as far as I can tell

What about `escape = pure . unTagged`? Then you no longer need to care about `s`.

qrpnxz:
I perhaps should have mentioned that not everything would be exported. You would not allow unwrapping `Tagged` normally. Nor allow running the effect except with `with` (can be named something else). And unwrapped versions of resource-using operations (e.g. `hTell :: Handle -> _ Integer`) would ideally not exist, or be part of an unsafe interface.

qrpnxz:
[...]

PR #98, opened 2022-09-09
STE showcase

arybczak:
As a demonstration for #93.

### DIFF
--- a/effectful/src/Effectful/STE.hs
+++ b/effectful/src/Effectful/STE.hs
@@ -1,0 +1,84 @@
+module Effectful.STE
+  ( STE
+
+  , MutVar
+  , runSTE
+  , newMutVar
+  , readMutVar
+  , writeMutVar
+  , atomicModifyMutVar
+  , modifyMutVar
+
+  , MutVarU
+  , newMutVarU
+  , readMutVarU
+  , writeMutVarU
+  , modifyMutVarU
+  ) where
+
+import Control.Monad.Primitive
+import Data.Coerce
+import Data.Primitive.Types
+import qualified Data.Primitive.MutVar as P
+import qualified Data.Primitive.PrimArray as P
+
+import Effectful
+import Effectful.Dispatch.Static
+import Effectful.Internal.Effect
+
+-- | Dirty trick for disambiguation of STE type parameters.
+instance {-# INCOHERENT #-} s1 ~ s2 => STE s1 :> (STE s2 : es) where
+  reifyIndex = 0
+
+type role STE nominal phantom phantom
+data STE s :: Effect
+type instance DispatchOf (STE s) = Static NoSideEffects
+data instance StaticRep (STE s) = STE
+
+runSTE :: (forall s. Eff (STE s : es) a) -> Eff es a
+runSTE m = evalStaticRep STE m
+
+----------------------------------------
+-- MutVar
+
+type role MutVar nominal representational
+newtype MutVar s a = MutVar (P.MutVar RealWorld a)
+  deriving Eq
+
+newMutVar :: STE s :> es => a -> Eff es (MutVar s a)
+newMutVar = coerce . unsafeEff_ . P.newMutVar
+
+readMutVar :: STE s :> es => MutVar s a -> Eff es a
+readMutVar = unsafeEff_ . P.readMutVar . coerce
+
+writeMutVar :: STE s :> es => MutVar s a -> a -> Eff es ()
+writeMutVar (MutVar mv) a = unsafeEff_ $ a `seq` P.writeMutVar mv a
+
+atomicModifyMutVar :: STE s :> es => MutVar s a -> (a -> (a, b)) -> Eff es b
+atomicModifyMutVar (MutVar mv) = unsafeEff_ . P.atomicModifyMutVar' mv
+
+modifyMutVar :: STE s :> es => MutVar s a -> (a -> a) -> Eff es ()
+modifyMutVar (MutVar mv) = unsafeEff_ . P.modifyMutVar' mv
+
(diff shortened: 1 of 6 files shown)


arybczak:
@qrpnxz I came up with a dirty trick to make the effect usable (with minor hassle) without `effectful-plugin`. I think it's worth adding other mutable structures from `primitive` and putting it into `effectful` proper.

effectful/src/Effectful/STE.hs
```diff
@@ -0,0 +1,84 @@
...
+
+import Effectful
+import Effectful.Dispatch.Static
+import Effectful.Internal.Effect
+
+-- | Dirty trick for disambiguation of STE type parameters.
+instance {-# INCOHERENT #-} s1 ~ s2 => STE s1 :> (STE s2 : es) where
+  reifyIndex = 0
+
+type role STE nominal phantom phantom
```
qrpnxz, line 33: I'm not sure what is appropriate role for the last two type arguments, but the `s` is absolutely phantom. How is `nominal` appropriate?

arybczak: This just mimics the API of `ST`, where `s` is `nominal`. I don't know if making it phantom has any consequences, so I left it like that.

For sure `MutVar` etc. need the `s` to be `nominal`, otherwise you can just cast it to `()` with `coerce` and pass it outside of the scope of `runSTE`.

qrpnxz: Indeed! Thanks.

qrpnxz:
Props on that trick! Hopefully it's applicable to #96. I must try it out.

I assume you have that custom `MutVar` type to avoid doing the `unsafeCoerce`, but I don't think that's really buying you much (if anything) in terms of safety, and it hurts interoperability with code using `primitive`'s `MutVar`. Even for the unboxed variant, which is not already in `primitive`, I would rather it be usable in any `PrimMonad` (automatic first-class compatibility with `Prim`, `ST`, `IO`, and more!), and then have a merely convenient `STE`-specialized wrapper API.

I'm surprised to not see here `primToSTE`, `stToSTE`, etc. that allow you to use *anything* from `primitive` (and other `ST`-only APIs) out of the gate. Maybe it's because this PR is "just a showcase", and so will be missing a lot, but it seems like the minimum API one would want.

I'd also recommend having `MutVar` code, `MutVarU` code, etc. in separate modules for qualified importing (and it just makes sense from an organization perspective).

arybczak:
> Maybe it's because this PR is "just a showcase", and so will be missing a lot, but it seems like the minimum API one would want.

Yes, that's the case. It's a demo, full version would have wrappers for all modules from `primitive`.

> I don't think that's really buying you much (if anything) in terms of safety, and it hurts interoperability with code using primitive's MutVar

The point here is to allow to use `MutVar` etc. within `runPureEff`. You get the interop with `primitive` via the `Prim` effect, but you cannot use it within `runPureEff` because of [this](https://github.com/haskell-effectful/effectful/issues/93#issuecomment-1237060602) issue.

I don't see how this can be done differently than providing `STE` effect with wrappers for stuff from `primitive`.

Which, now that you put it like that, might in fact not be worth it.

qrpnxz:
> The point here is to allow to use MutVar etc. within runPureEff. [...] I don't see how this can be done differently than providing STE effect with wrappers for stuff from primitive.

The way I'm currently using [effectful-st] looks like this: `stToSTE (P.newMutVar a)`. Later I do `runSTE` and bam, we can use `runPureEff` (as long as we didn't do anything else that required `IOE`).

It would certainly be convenient to have wrappers for `newMutVar` and such from `primitive` (it adds no dependency, after all), but it shouldn't be the only thing available. With that function we can use `vector`, `hashtables`, `STRef` and more without pulling them in, wrapping them, etc.

[effectful-st]: https://hackage.haskell.org/package/effectful-st-0.0.0.0/docs/Effectful-ST.html

arybczak:
> The way I'm currently using [effectful-st](https://hackage.haskell.org/package/effectful-st-0.0.0.0/docs/Effectful-ST.html) looks like this: stToSTE (P.newMutVar a)

Don't you have to wrap every function that manipulates this `MutVar` with `stToSTE` later on? If so, it's not very ergonomic, isn't it?

qrpnxz:
It has the ergonomics of `liftIO` basically :) In my code, I didn't bother to wrap anything. I sprinkle `stToSTE` in the few places I'm modifying or reading state. I have a 100 line function here that uses it like five times. Two of those times I call it on an entire block like `stToSTE $ do {...}`, which would otherwise be 1-3 uses of `stToSTE` at particular use sites. For me personally, for this example, I didn't feel bad about having to use `stToSTE` whatsoever. And considering the alternative was `Prim`, I'm, on the contrary, elated that I got such a simple solution.

qrpnxz:
A worse-case scenario would be having to interleave a bunch of different `STE` with actions from other `Effects`:
```hs
catch ...
stToSTE (P.newMutVar ...)
readFile ...
stToSTE (GMV.new ...)
...
```
which I don't find too bad, but otherwise you can do this:
```hs
...
stToSTE $ do
  ...
...
```
or this
```hs
let func = stToSTE . P.func
...
```
or if `STE` is not that common, just `stToSTE` at the use site.

Lifting in general is not too big a deal. We do it all the time with `fmap`, `pure`, etc. `lift` from `MonadTrans` is annoying for different reasons (`lift . lift . lift . lift . lift` "...was it five times?").

qrpnxz:
I actually had to call `pure` the same number of times I had to call `stToSTE`, and I didn't think twice about that.

arybczak:
Fair enough. Let's close this then.

Issue #99, opened 2022-09-16
NonDet handler that executes multiple branches


arybczak:
It would be nice to have, but I tried to naively add it using freshly merged delimited continuations (upcoming GHC 9.6) and it doesn't really work. Details here: https://github.com/re-xyr/speff/issues/1#issuecomment-1247981148

arybczak:
Considering that this requires (delimited) continuations and continuations are not compatible with the `MonadUnliftIO` instance (which I'm not willing to drop), this will most likely never happen.

PR #100, opened 2022-10-03
Add Alternative and MonadPlus instances for Eff

mmhat:
This PR adds two new instances `Alternative (Eff es)` and `MonadPlus (Eff es)`. In the case of alternative `mx <|> my` we proceed as follows:
 1. For the left branch `mx`: Create a copy of the current environment, run the branch with that environment, and on success overwrite the current environment with the copy.
 2. If the left branch fails, run the right one.
 
`empty` throws a new exception that was added to the library: `data AlternativeException = EmptyAlternative`. Note that this not as safe as the approach chosen for `MonadFail`: Here, using the instance requires the `Fail` effect to be in scope that handles the failure. The idea behind `AlternativeException` was that it should be easy to write something like `foo <|> bar <|> pure default` without having to deal with the empty case explicitly. Whether this is a good idea is of course up for debate :slightly_smiling_face:

### DIFF
--- a/effectful-core/src/Effectful.hs
+++ b/effectful-core/src/Effectful.hs
@@ -53,6 +53,9 @@ module Effectful
   , inject
   , Subset
 
+    -- * Exceptions
+  , AlternativeException(..)
+
     -- * Re-exports
   , MonadIO(..)
   , MonadUnliftIO(..)
--- a/effectful-core/src/Effectful/Internal/Env.hs
+++ b/effectful-core/src/Effectful/Internal/Env.hs
@@ -18,6 +18,7 @@ module Effectful.Internal.Env
     -- * Operations
   , emptyEnv
   , cloneEnv
+  , overwriteEnvWith
   , sizeEnv
   , tailEnv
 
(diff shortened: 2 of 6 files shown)


mmhat:
@arybczak I left some todos in `Effectful.Internal.Env` (https://github.com/haskell-effectful/effectful/blob/1aceed32b3fe5fc888c61c82b19b6acb62b6c468/effectful-core/src/Effectful/Internal/Env.hs#L164) as I was not sure if overwriting an environment needs something more than copying the storage. I also included the `interpose` tests to check that and it appears to work. None the less, can you double-check that this is sane?

Also, I didn't verify the `MonadPlus` laws (yet), but I think those should be fine.

arybczak:
The `Alternative` instance usually would delegate to the dynamic `NonDet` effect (see how it's done in `eff`).

The "proper" `NonDet` interpreter that executes all branches currently cannot be written (see #99), but you can provide an interpretation that will do what you do in this PR. Then, perhaps one day a proper one can be provided.

mmhat:
Closing in favour of #102.

Issue #112, opened 2022-11-17
withSeqEffToIO and withConcEffToIO functions


Nexmean:
It would be nice to have functions for unlifting Eff monad with chosen strategy, but without change of unlift strategy for all computations

arybczak:
Fixed by #113. Thanks for the PR :slightly_smiling_face:

PR #113, opened 2022-11-18
Add withSeqEffToIO and withConcEffToIO functions

Nexmean:
address issue https://github.com/haskell-effectful/effectful/issues/112

### DIFF
--- a/effectful-core/src/Effectful/Internal/Monad.hs
+++ b/effectful-core/src/Effectful/Internal/Monad.hs
@@ -43,6 +43,8 @@ module Effectful.Internal.Monad
   , unliftStrategy
   , withUnliftStrategy
   , withEffToIO
+  , withSeqEffToIO
+  , withConcEffToIO
 
   -- ** Low-level unlifts
   , seqUnliftIO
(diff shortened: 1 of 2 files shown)


arybczak:
Thanks. I'd prefer the function with a signature

```
???
  :: (HasCallStack, IOE :> es)
  => UnliftStrategy
  -> ((forall r. Eff es r -> IO r) -> IO a)
  -> Eff es a
```

though for consistency. But I'm not sure about its name :thinking: `withEffToIO'` in theory works, but I really dislike function variations with ticks since they're non-descriptive.

Nexmean:
I think that common case for this functions is to implement some effect that need to provide callback to library that use IO. So there you don't need to send to your functions UnliftStrategy as argument. And it will be cleaner to call just `withSeqEffToIO` than call `withUnliftStrategyEffToIO SeqUnlift`.

arybczak:
You seem to have a particular use-case in mind, can you show me?

Nexmean:
For example, I use `hw-kafka-client` and I need to provide log callback to it:
```haskell
getKafkaLoggerCallback :: (IOE :> es, Logger :> es) => Eff es Kafka.Callback
getKafkaLoggerCallback = withConcEffToIO Ephemeral Unlimited \effToIO ->
  pure $ Kafka.logCallback \level facility message ->
    effToIO
    $ addLoggingNamespace "kafka"
    $ addLoggingContext (sl "kafka_facility" facility)
    $ logWithLevel (fromKafkaLogLevel level)
    $ fromString message
```
Or when I use `servant-client` I need to provide callback for logging request and response, but with `SeqUnlift` strategy.

Or when I use `Resource` effect I need logging at acquiring and closing too.

effectful-core/src/Effectful/Internal/Monad.hs
```diff
@@ -179,9 +181,28 @@ withEffToIO
   -- ^ Continuation with the unlifting function in scope.
   -> Eff es a
 withEffToIO f = unliftStrategy >>= \case
-  SeqUnlift -> unsafeEff $ \es -> seqUnliftIO es f
+  SeqUnlift      -> unsafeEff $ \es -> seqUnliftIO es f
   ConcUnlift p b -> unsafeEff $ \es -> concUnliftIO es p b f
 
+-- | Create an unlifting function with 'SeqUnlift' strategy.
```
arybczak, line 187:
```suggestion
-- | Create an unlifting function with the 'SeqUnlift' strategy.
```

effectful-core/src/Effectful/Internal/Monad.hs
```diff
@@ -179,9 +181,28 @@ withEffToIO
...
+withSeqEffToIO
+  :: (HasCallStack, IOE :> es)
+  => ((forall r. Eff es r -> IO r) -> IO a)
+  -- ^ Continuation with the unlifting function in scope.
+  -> Eff es a
+withSeqEffToIO f = unsafeEff $ \es -> seqUnliftIO es f
+
+-- | Create an unlifting function with 'ConcUnlift' strategy.
```
arybczak, line 195:
```suggestion
-- | Create an unlifting function with the 'ConcUnlift' strategy.
```

arybczak (commented):
Ok, that looks reasonable :+1:

PR #131, opened 2023-01-22
Test with GHC 9.6

arybczak:


arybczak:
Sigh. Ok, let's try again later.

PR #135, opened 2023-01-31
Add support for turning an effect handler into an effectul operation

### LINKED ISSUE #134: [Question] Scoped effects in effectful
Hi!

Does `effectful` support something like `polysemy`'s [Scoped](https://www.tweag.io/blog/2022-01-05-polysemy-scoped/)?

If not, is there a way to handle having state transactions *in* my business logic in a type safe way, i.e.
* I should not be able to use my "stateful effect" outside a transaction
* I should not have to leak the implementation of the transaction (e.g. acquiring an sql connection from a pool)

?

arybczak:
Fixes #134.

@googleson78 does that work for you?

FWIW if it ends up getting merged, I'll probably rename it to something else, because it's really about turning an interpreter into an effect, so `Scoped` is a bit weird name for this. EDIT: though I'm not sure what the more appropriate name would be.

Provided `runScoped` should be enough for all use cases. Runners in `polysemy` have insane types and I don't know what are the differences between them, but `runScopedNew` supposedly subsumes them and IIUC my `runScoped` is a more readable version of it.

~~Also, looks like it might be somewhat annoying to use without the plugin if you make use of an effect with parameters.~~ Nah, the current order of type parameters makes it alright.

### DIFF
--- a/effectful-core/CHANGELOG.md
+++ b/effectful-core/CHANGELOG.md
@@ -3,6 +3,8 @@
 * Make `withEffToIO` take an explicit unlifting strategy for the sake of
   consistency with unlifting functions from `Effectful.Dispatch.Dynamic` and
   easier to understand API.
+* Add support for turning an effect handler into an effectful operation via the
+  `Reified` effect.
 
 # effectful-core-2.2.2.2 (2023-03-13)
 * Allow `inject` to turn a monomorphic effect stack into a polymorphic one.
--- a/effectful-core/effectful-core.cabal
+++ b/effectful-core/effectful-core.cabal
@@ -89,6 +89,7 @@ library
                      Effectful.Prim
                      Effectful.Reader.Dynamic
                      Effectful.Reader.Static
+                     Effectful.Reified
                      Effectful.State.Dynamic
                      Effectful.State.Static.Local
                      Effectful.State.Static.Shared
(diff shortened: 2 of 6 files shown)


arybczak:
@googleson78 do you plan on having a look?

googleson78:
Ah, sorry for going afk. I was mostly just curious originally :sweat_smile: . The examples and types (obviously) look about right, I'll be sure to try this out whenever I get around to playing with `effectful`. Without doing so, I don't think I have anything useful to comment.

arybczak:
I'd like to finish and merge this PR since I have a use case, but I still don't particularly like the name `Scoped`.

Ideal would be `Handler`/`handle`/`runHandler`, but that clashes with stuff from `Control.Exception`.

Second best IMHO is `Runner`/`run`/`runRunner`, though `run` might also clash and `runRunner` looks a bit weird.

Perhaps `Provider`/`provide`/`runProvider` :thinking: 

Thoughts?

ocharles:
I can't think of a much better name, but `Scoped` might not be awful if that's the nomenclature `polysemy` is using. Other than that, the best I can think of is `DynamicInterpreter` - but that's ambiguous given that `interpret` is already in `Effectful.Dispatch.Dynamic`

arybczak:
@ocharles thanks, I completely forgot about `Reified`, that sounds good to me :slightly_smiling_face:

PR #138, opened 2023-02-03
Add support for labeled effects

arybczak:
Proof of concept for support of having multiple effects of the same type in scope.

### DIFF
--- a/effectful-core/CHANGELOG.md
+++ b/effectful-core/CHANGELOG.md
@@ -7,6 +7,8 @@
   `Provider` effect.
 * Add `runErrorWith` and `runErrorNoCallStackWith` to `Effectful.Error.Dynamic`
   and `Effectful.Error.Static`.
+* Add support for having multiple effects of the same type in scope via the
+  `Labeled` effect.
 
 # effectful-core-2.2.2.2 (2023-03-13)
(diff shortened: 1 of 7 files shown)


arybczak:
Turns out `labeled` is too clever and thus doesn't work properly if `e :> es` in already in the context.

It needs to be changed to

```haskell
labeled
  :: forall label e es a
   . Labeled label e :> es
  => Eff (e : es) a
  -- ^ An action using the effect.
  -> Eff es a
```

but then it will have to copy `envRefs` every time it's called :( It's a `PrimArray`, but still. Oh well, I'll benchmark and see how much of a problem this is.

smunix:
Any chance this will get merged soon? I've got a few uses where I need an effect to be in scope multiple times.

smunix:
> @smunix I'll merge this soon. I rebased over current master, feel free to use this branch for now with cabal's `source-repository-package` or something slightly_smiling_face

Branch adopted! It works wonders, thank you!

johnhampton:
👋 Is the plan still to merge this PR? I don't mind using the branch, but I wanted to make sure that this PR isn't being abandoned and that labelled effects will eventually make it into the library.

arybczak:
Feel free to use the branch. It'll end up in master at some point, I just didn't have time recently to clean up the documentation.

PR #174, opened 2023-07-31
Test with GHC 9.8

arybczak:


arybczak:
Never mind, release metadata in ghcup is busted.

Kleidukos:
A bit late to the party but yes: https://github.com/haskell/ghcup-metadata/issues/102

PR #181, opened 2023-09-10
Update CI

arybczak:


arybczak:
@mmhat I removed bounds for `tasty-bench`, it looks like the failure you got was the result of [no upper `tasty` bounds on older versions of tasty-bench](https://hackage.haskell.org/package/tasty-bench-0.3.3/revisions/).

PR #213, opened 2024-03-30
Add Effectful.Coroutine

tomjaguarpaw:
Would you be open to adding this sort of `Coroutine` effect, which provides access to stream processing, matching some of the functionality, but not all, of `pipes`, `conduit` and `streaming`? Even though these are not first-class coroutines (they cannot be paused and passed around) they are extremely convenient in practice.

I am completely open to reworking the PR if you think there's a better implementation or interface.

Note that, because they are not first-class, this PR does not invalidate [the claim in the README](https://github.com/haskell-effectful/effectful?tab=readme-ov-file#any-downsides) that `effectful` cannot provide a `Coroutine` effect. Perhaps the effect here should be named differently. I'm not sure.

Credit to @Lysxia for the original implementation

### DIFF
--- a/effectful-core/effectful-core.cabal
+++ b/effectful-core/effectful-core.cabal
@@ -75,6 +75,7 @@ library
       c-sources:     cbits/utils.c
 
     exposed-modules: Effectful
+                     Effectful.Coroutine
                      Effectful.Dispatch.Dynamic
                      Effectful.Dispatch.Static
                      Effectful.Dispatch.Static.Primitive
--- a/effectful-core/src/Effectful/Coroutine.hs
+++ b/effectful-core/src/Effectful/Coroutine.hs
@@ -1,0 +1,84 @@
+-- | 'Coroutine' gives you access to stream processing, similar to the
+-- @conduit@, @pipes@ or @streaming@ libraries.  'Coroutine' is not a
+-- "first class" coroutine: a coroutine computation can't be suspended
+-- or passed around.
+
+module Effectful.Coroutine
+  ( -- * Effect
+    Coroutine (..),
+
+    -- ** Handlers
+
+    forEach,
+    yieldToList,
+    yieldToReverseList,
+
+    -- ** Operations
+
+    call,
+    yield,
+    inFoldable,
+  )
+where
+
+import Data.Foldable (traverse_)
+import Effectful
+  ( Dispatch (Dynamic),
+    DispatchOf,
+    Eff,
+    Effect,
+    type (:>),
+  )
+import Effectful.Dispatch.Dynamic (interpret, reinterpret, send)
+import Effectful.State.Static.Local (modify, runState)
+
+data Coroutine a b :: Effect where
+  Call :: a -> Coroutine a b m b
+
+type instance DispatchOf (Coroutine a b) = Dynamic
+
+-- | Call the 'Coroutine'.
+call :: forall a b es. (Coroutine a b :> es) => a -> Eff es b
+call a = send (Call a)
+
+-- | The same as 'call', but specialised to the case when the call
+-- returns @()@, for better type inference.
+yield :: forall a es. (Coroutine a () :> es) => a -> Eff es ()
+yield = call
+
+-- | Handle the coroutine.
+--
+-- @
+-- runPureEff $ execState 0 $ do
+--     forEach \@Int (inFoldable \@Int [0 .. 5]) $ \\i -> do
+--       modify (+ i)
+-- 15
+-- @
+forEach :: forall a b es r. Eff (Coroutine a b ': es) r -> (a -> Eff es b) -> Eff es r
+forEach caller call_ = handle caller
+  where
+    handle = interpret $ \_ -> \case
+      Call a -> call_ a
+
+-- | Yield all the elements of the @Foldable@.
+inFoldable :: forall a t es. (Foldable t, Coroutine a () :> es) => t a -> Eff es ()
+inFoldable = traverse_ yield -- can't use unannotated `call`; type inference fail
+
+-- | Handle the coroutine by gathering the yielded values into a list.
+--
+-- @
+-- runPureEff $ yieldToList $ do
+--   for_ [0 .. 3] $ \\i -> do
+--     yield \@Int i
+--     yield (i * 10)
+-- ([0,0,1,10,2,20,3,30],())
+-- @
+yieldToList :: forall a es r. Eff (Coroutine a () ': es) r -> Eff es ([a], r)
+yieldToList =
+  (fmap . fmap) (\(as, r) -> (reverse as, r)) yieldToReverseList
+
+yieldToReverseList :: forall a es r. Eff (Coroutine a () ': es) r -> Eff es ([a], r)
+yieldToReverseList = reinterpret @(Coroutine a ()) collect $ \_ -> \case
+  Call x -> modify (x :)
+  where
+    collect = fmap (\(r, xs) -> (xs, r)) . runState []


arybczak:
> Would you be open to adding this sort of Coroutine effect, which provides access to stream processing, matching some of the functionality, but not all, of pipes, conduit and streaming?

Not particularly. If you need streaming capabilities, you should use one of the dedicated streaming libraries. AFAIK they integrate very well with `effectful` and their authors spent a ton of time making them performant and their API comprehensive. I don't think providing an ad-hoc, partial version of streaming API here is any good, it will just make people choose the wrong path in their programming.

On the other hand, if you had any issues with using `effectful` with streaming libraries, I'd like to hear about it, perhaps the situation can be improved in this regard.

tomjaguarpaw:
> On the other hand, if you had any issues with using `effectful` with streaming libraries, I'd like to hear about it, perhaps the situation can be improved in this regard.

Sure, it's not a problem for me personally because I can just use [Bluefin](https://hackage.haskell.org/package/bluefin) instead of `effectful`, but I'll write it up in case it will help anyone who is using `effectful`.

By way of example how would you implement `enumerate` using `effectful`? It implements the same kind of thing as [Python's `enumerate`](https://docs.python.org/3/library/functions.html#enumerate), i.e. for each element of the input stream it yields a tuple of that element paired with a count of the elements seen so far.

Here it is using `StateT`:

```.hs
import Control.Monad.Trans (lift)
import Control.Monad.Trans.State (evalStateT, get, put)
import Streaming (hoist)
import qualified Streaming.Prelude as P
import Streaming.Prelude (Stream, Of, yield, each)

enumerate :: Monad m => Stream (Of a) m r -> Stream (Of (Int, a)) m r
enumerate s =
  flip evalStateT 0 $ do
    flip P.mapM_ (hoist (lift . lift) s) $ \a -> do
      n' <- get
      lift (yield (n', a))
      put $! (n' + 1)

-- ghci> example
-- (0,"foo")
-- (1,"bar")
-- (2,"baz")
example :: IO ()
example =
  P.mapM_ print (enumerate (each ["foo", "bar", "baz"]))
```

Those `hoist`s and `lift`s are very ugly, and supposedly what effect systems are supposed to help us with. So can the situation be improved using `effectful`. No, I don't think so. Here's a partial attempt using `effectful`:

```.hs
import Effectful (Eff)
import Effectful.State.Dynamic (evalStateLocal, get, put)
import qualified Streaming.Prelude as P
import Streaming.Prelude (Stream, Of, yield, each)

enumerateEf ::
  Stream (Of a) (Eff es) r ->
  Stream (Of (Int, a)) (Eff es) r
enumerateEf s =
  evalStateLocal 0 $ do
    flip P.mapM_ (_ s) $ \a -> do
      n' <- get
      _ (yield (n', a))
      put $! (n' + 1)
```

But I don't think there's any way to complete this attempt. When we use `P.mapM_` we can iterate over the input stream in any `Monad` `m`. But the closest handler is `evalStateLocal`, so there cannot be any outer layer of `Stream` on `m`. `m` must be just `Eff es`. So there is no way that we can propagate the `Stream` effect out, and we are stuck. By contrast, here is how it looks in Bluefin, with the sort of coroutines I'm proposing in this PR:

```.hs
{-# LANGUAGE RankNTypes #-}
{-# LANGUAGE TypeOperators #-}

import Bluefin.Eff (Eff, runEff, (:&), (:>))
import Bluefin.IO (effIO)
import Bluefin.State (evalState, get, put)
import Bluefin.Stream (Stream, inFoldable, forEach, yield)

enumerateBf ::
  e :> es =>
  (forall e st. Stream a e -> Eff (e :& st :& es) r) ->
  Stream (Int, a) e ->
  Eff es r
enumerateBf s y =
  evalState 0 $ \state -> do
    forEach s $ \a -> do
      n' <- get state
      yield y (n', a)
      put state (n' + 1)

-- ghci> exampleBf
-- (0,"foo")
-- (1,"bar")
-- (2,"baz")
exampleBf :: IO ()
exampleBf = runEff $ \io -> do
  forEach (enumerateBf (inFoldable ["foo", "bar", "baz"])) $ \t -> do
    effIO io (print t)
```

No `hoist`s, no `lift`s, just direct style code (quite similar to what we would have written in Python). Because Bluefin has a coroutine effect it doesn't matter that the state handler is the closest one. We're free to propagate coroutine effects through it.

I can imagine a few possible responses to this

* *Just use a streaming library that has this operation built in*

  Sure, but this is just a simple example to demonstrate the problem. The point is that a streaming transformer on top of `effectful` is extremely limited in what it can do using `effectful`, because the streaming transformer is always on the outside and must *always* be handled first.

  (By the way, this is the dual problem that we have with `StateT s IO`, where `IO` is on the inside and must always be handled last, i.e. you can't preserve the state when an IO exception is thrown.)

* *Just use `StateT` anyway*

  But [`effectful` explicitly warns against `StateT`](https://github.com/haskell-effectful/effectful/blob/a235b074cd68d770cbe732dc2a65067ff0c754f5/transformers.md), so that would be strange.

* *Just find another way of implementing it, for example a recursive definition of `enumerate`*

  Sure, but firstly this is just a simple example; a realistic example will be much harder to write in another style. Secondly, effect systems are supposed to avoid hand-coding of effects anyway, so this seems to defeat the point of using an effect system.

----

To reiterate, personally I don't mind whether something like this makes it into `effectful` because I'll be using Bluefin anyway, but I think lightweight streaming abstractions are *extremely* useful in practice, and something like this could be very helpful to `effectful` users. But you know your users better than me, of course.

tomjaguarpaw:
I don't know your policy on closing PRs, so I'll close this myself in due course unless I hear otherwise.

arybczak:
> By way of example how would you implement enumerate using effectful?

I don't know about `streaming` (I find its API confusing), but e.g. `conduit` has [mapAccum](https://hackage.haskell.org/package/conduit-1.3.5/docs/Data-Conduit-List.html#v:mapAccum) for stateful streaming. I'm sure other libraries have similar transformations.

> Just use a streaming library that has this operation built in
>
> Sure, but this is just a simple example to demonstrate the problem. The point is that a streaming transformer on top of effectful is extremely limited in what it can do using effectful, because the streaming transformer is always on the outside and must always be handled first.

That's true, you can't interleave running effects and streaming, you need to run all the effects you will need in your stream up front. I'd like to see a "real-world" example where this is a problem.

Btw, this effect shouldn't really be called `Coroutine`, because it's not. It's just turning a function call into an effectful operation, so

```haskell
-- | Provide a way to call an effectful function.
data Function a b :: Effect where
  Call :: a -> Function a b m b

type instance DispatchOf (Function a b) = Dynamic

call :: (HasCallStack, Function a b :> es) => a -> Eff es b
call = send . Call

call_ :: (HasCallStack, Function a () :> es) => a -> Eff es ()
call_ = call
```

would be reasonable.

But I'm not going to add it to `effectful-core` because I have no idea whether the runners are reasonable (my first impression is that most runners and examples are weird, I might be wrong though).

Considering that this is a very simple ordinary effect, nothing prevents anyone from writing a library `effectful-streaming` or something and experimenting with this interface there.

I'm conservative about adding stuff to the main library that I don't have a good intuition about, because once it's there, it's hard to change it.

tomjaguarpaw:
> conduit has [mapAccum](https://hackage.haskell.org/package/conduit-1.3.5/docs/Data-Conduit-List.html#v:mapAccum) for stateful streaming

Right, that's my suggested answer 3, "*Just use a streaming library that has this operation built in*". 

> I'd like to see a "real-world" example where this is a problem.

What could be more "real-world" than using an effect system to handle effects? Here's a more involved example. It reads stream elements and yields their sum, until the sum reaches 10.

```.hs
import Control.Monad (when)
import Control.Monad.Trans (lift)
import Control.Monad.Trans.State (evalStateT, get, put)
import Control.Monad.Trans.Except (runExceptT, throwE)
import Streaming (hoist)
import qualified Streaming.Prelude as P
import Streaming.Prelude (Stream, Of, yield, each)

emitSum :: Monad m => Stream (Of Int) m r -> Stream (Of Int) m ()
emitSum s = do
  runExceptT $ do
    flip evalStateT 0 $ do
      flip P.mapM_ (hoist (lift . lift . lift) s) $ \a -> do
        n' <- get
        let next = n' + a
        (lift . lift) (yield next)
        when (next >= 10) $
          lift (throwE ())
        put $! next

  pure ()

-- ghci> example
-- 1
-- 3
-- 6
-- 10
example :: IO ()
example =
  P.mapM_ print (emitSum (each [1..100]))
```

The `hoist`s and `lift`s are even worse with two local effects! By contrast here's the Bluefin version:

```.hs

{-# LANGUAGE RankNTypes #-}
{-# LANGUAGE TypeOperators #-}

import Bluefin.Eff (Eff, runEff, (:&), (:>))
import Bluefin.IO (effIO)
import Bluefin.State (evalState, get, put)
import Bluefin.Stream (Stream, inFoldable, forEach, yield)
import Bluefin.Jump (jumpTo, withJump)
import Control.Monad (when)

enumerateBf ::
  e :> es =>
  (forall e st j. Stream Int e -> Eff (e :& st :& j :& es) ()) ->
  Stream Int e ->
  Eff es ()
enumerateBf s y = do
  withJump $ \j -> do
    evalState 0 $ \state -> do
      forEach s $ \a -> do
        n' <- get state
        let next = n' + a
        yield y next
        when (next >= 10) $
          jumpTo j
        put state next

-- ghci> exampleBf
-- 1
-- 3
-- 6
-- 10
exampleBf :: IO ()
exampleBf = runEff $ \io -> do
  forEach (enumerateBf (inFoldable [1..100])) $ \t -> do
    effIO io (print t)
```

No `hoist`s or `lift`s, just as we expect from an effect system.

If you want to think even more real world, then let's interpret what I'm asking for in terms of data processing: I'm asking to be able to ingest a stream of data, transform it using some local effects, and emit one or more streams of transformed data in response. It doesn't get more real world than that.

Perhaps it's because I used Python before Haskell, but I'm convinced that access to lightweight streams/generators/iterators is a game changer for clarity of programming. To be truly lightweight they need to be integrated into an effect system and to be able to interact in a native way with everything that effect system provides.

> this effect shouldn't really be called `Coroutine`, because it's not

Do you have a source with a reliable definition of "coroutine" that rules out this nomenclature? [Wikipedia suggests](https://en.wikipedia.org/wiki/Coroutine#Definition_and_types) that it could be a "second-class stackful coroutine". I'm also open to other names, but not "function". It *is* a function of course[^1], but calling it "function" does not accurately describe its role. One would use `yield` with the intuition that one is "yielding" to a handler higher up the stack, not "calling" the handler.

> I'm conservative about adding stuff to the main library that I don't have a good intuition about, because once it's there, it's hard to change it.

That's understandable, and I appreciate that policy. I'll close.

[^1]: This is functional programming. Everything is a function.

ocharles:
What I've found hardest to understand for your argument is when this example crops up but we can't easily express it in what I consider a more idiomatic Haskell style. For example, your "accumluate-up-to" example is written very imperatively - the alternative is to just use `scanl` and `takeWhile`. I know you sort of touched on this by saying that it could be written recursive style but that's besides the point, but this suggestion is neither recursive nor imperative - it's (imo) idiomatic Haskell. What I think would help me understand the desire for this effect more would be to see how it's useful within idiomatic Haskell. Naturally we're already off to a rocky start here because no one can even establish what idiomatic Haskell even is!

tomjaguarpaw:
I'm not sure I understand what you're saying here, but I can clarify a few points:

* What I'm suggesting is indeed non-idiomatic, if you call the `yield`-oriented style of stream programming non-idiomatic (I use the `streaming` library in that style, for example -- others undoubtedly use it in a different style)
* The style is indeed imperative. I don't think that should be at odds with "functional" style. See [a Discourse thread on functional style](https://discourse.haskell.org/t/beautiful-functional-programming/7411/52) for more of my thoughts on that.
* One of my aims with Bluefin is to make this imperative style more convenient in Haskell, and to demonstrate that it is an effective way of programming that doesn't subvert any of the benefits of Haskell that we love.

michaelpj:
Something like this does seem quite useful. I'm intrigued by this comment 

> But I'm not going to add it to effectful-core because I have no idea whether the runners are reasonable (my first impression is that most runners and examples are weird, I might be wrong though).

I would be tempted to write this locally and use it myself, but I am concerned that it is in fact weird! Do we know what kind of weirdness we might expect?

tomjaguarpaw:
I don't know of anything weird about it. I've been using the Bluefin equivalent for a while, it's been very useful and I haven't experienced any problems.

arybczak:
> I would be tempted to write this locally and use it myself, but I am concerned that it is in fact weird! Do we know what kind of weirdness we might expect?

To clarify, there is no operational weirdness here as it's "just" an ordinary effect :slightly_smiling_face:  What I meant is "I've never seen such code, I've never needed such code and examples can be written in a more natural (for the lack of a better word) style, therefore I'm not convinced that it's a good addition to the core library". FTR @ocharles sentiment above also resonates with me.

tomjaguarpaw:
Here's an example that I find really compelling. Using Bluefin [`Coroutine`](https://hackage.haskell.org/package/bluefin-0.0.7.0/docs/Bluefin-Coroutine.html)s, we can do a similar job to `pipes` or `conduit`, but unlike those libraries we can terminate early and nonetheless promptly free resources without having to use a special abstraction (`ResourceT` or `SafeT`). Bluefin's standard [`bracket`](https://hackage.haskell.org/package/bluefin-0.0.7.0/docs/Bluefin-Eff.html#v:bracket) works perfectly well. In the example below I use `bracket` inside a "producer"[^2] (that produces indefinitely) and it nonetheless promptly frees its "resource"[^1] when the consumer terminates, without having to remember to tightly wrap a `runResourceT`/`runSafeT` around ... something. It just does the right job in all circumstances.

effectful could have this too.

```
-- Acquiring resource
-- 1
-- 2
-- 3
-- 4
-- 5
-- Releasing resource
-- Finishing
promptCoroutine :: IO ()
promptCoroutine = runEff $ \io -> do
  -- receiveStream connects a consumer to a producer
  receiveStream
    -- Like a pipes Consumer.  Prints the first five elements it
    -- receives.
    ( \r -> for_ [1 :: Int .. 5] $ \_ -> do
        v <- yieldCoroutine r ()
        effIO io (print v)
    )
    -- Like a pipes Producer. Yields successive integers indefinitely.
    -- Unlike in pipes, we can simply use Bluefin's standard bracket
    -- for prompt release of a resource
    ( \y ->
        bracket
          (effIO io (putStrLn "Acquiring resource"))
          (\_ -> effIO io (putStrLn "Releasing resource"))
          (\_ -> for_ [1 :: Int ..] $ \i -> yield y i)
    )
  effIO io (putStrLn "Finishing")
```

[^1]: in this example there is no actual resource

[^2]: Bluefin calls "producers" `Stream`s

arybczak:
> effectful could have this too

I'll wait for the `effectful-streaming` library then ;)

tomjaguarpaw:
> Here's an example that I find really compelling. Using Bluefin [Coroutine](https://hackage.haskell.org/package/bluefin-0.0.7.0/docs/Bluefin-Coroutine.html)s, we can do a similar job to pipes or conduit, but unlike those libraries we can terminate early and nonetheless promptly free resources

In fact, it's even worse for pipes and conduit. They don't actually finalize promptly, only "promptish". I wrote an article about this: [Bluefin streams finalize promptly](https://h2.jaguarpaw.co.uk/posts/bluefin-streams-finalize-promptly/).

PR #216, opened 2024-04-17
Rename localHandle to localLend and add localBorrow

arybczak:
@ocharles this renames the helper we discussed in #195 and adds one that works in the opposite direction (just fyi).


ocharles:
Nice names! I like these. I didn't end up using your last set of names (instead I passed a specific `Eff es -> Eff es'` function over), but when this reaches Hackage I'll switch over :+1: I can build from Git if you want me to test it before then

arybczak:
> I can build from Git if you want me to test it before then

If you have a moment for this, it would be great.

PR #226, opened 2024-07-03
Get haddock for makeEffect from the declaration

### LINKED ISSUE #225: Better haddock from `makeEffect`
At the moment `makeEffect` just inserts haddock like this:
```
Perform the effect <constructor name>.
```
However, it would be nice to be able to give users of the effect custom haddock. One option would be to take the haddock of the constructor, if there is any, and use that instead. I think that would give a fairly natural way of documenting effects?

michaelpj:
Fixes #225

I went ahead and put together a PR, since it's very simple.

I can't see a good way to test this, but you can verify by opening the test file in HLS, adding some references to the sending functions, and checking that they have the right haddock in the hover.


arybczak:
Thanks 🙂

PR #234, opened 2024-08-23
Generalize localLend/localBorrow to take a list of effects

arybczak:
@ocharles this will help with #233.


ocharles:
Thanks @arybczak ! I haven't had a chance to test this yet but the change looks good

PR #241, opened 2024-09-08
Heftia countdown benchmarks

arybczak:
```
package effectful
  flags: +benchmark-foreign-libraries
``` 

in `cabal.project.local` and `cabal run bench -- --pattern '$2 == "countdown" && $3 == "1000"' --svg bench_countdown_1000.svg` to reproduce (GHC 9.2.8).

bench_countdown_1000

I'm not merging this because these packages have way too restrictive bounds and pull in tons of deps, I was just curious.

### DIFF
--- a/effectful-th/effectful-th.cabal
+++ b/effectful-th/effectful-th.cabal
@@ -62,7 +62,7 @@ library
                     , effectful-core      >= 1.0.0.0   && < 3.0.0.0
                     , exceptions          >= 0.10.4
                     , template-haskell    >= 2.16      && < 2.23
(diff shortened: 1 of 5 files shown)


ymdryo:
Without this benchmark, I wouldn't have noticed the unexpectedly poor performance. Thank you.

The performance has now improved in v0.4 and is on par with `freer-simple`.
https://github.com/sayo-hs/heftia/blob/v0.4.0/benchmark/performance.md

If you're interested, please try running benchmarks there as well.
https://github.com/sayo-hs/heftia/blob/v0.4.0/heftia-effects/bench/BenchCountdown.hs

PR #257, opened 2024-10-14
Base Effectful.Exception on Control.Exception

### LINKED ISSUE #254: Effectful.Exception is subtly broken
Turns out that simply re-exporting functions from `safe-exception` was a mistake.

1. Functions like ~~`finally`~~ `onException` are subtly broken when used in transformers like `ExceptT` (it's highly unlikely someone will use `Effectful.Exception` to do this, but it's a possibility and I don't like that, gotta specialize them to the `Eff` monad).

2. `catch` from safe-exceptions behaves differently than `catch` from the `MonadCatch (Eff es)` instance :confused: 

I think the fix is to specialize functions to `Eff` for (1) and to leave `catch` with original semantics of the `exceptions` package, but provide helper functions like `catchAnySync` that catches all sync exceptions for (2) because this is the most common scenario where async exceptions are caught and swallowed by accident.

@ocharles @mmhat FYI since you participated in #248.

arybczak:
Fixes #254.

### DIFF
--- a/effectful-core/src/Effectful/Exception.hs
+++ b/effectful-core/src/Effectful/Exception.hs
@@ -5,7 +5,7 @@
 -- well as several utility functions for convenience.
 --
 -- /Note:/ the 'Eff' monad provides instances for 'C.MonadThrow', 'C.MonadCatch'
--- and 'C.MonadMask', so any code that uses them will also work.
+-- and 'C.MonadMask', so any existing code that uses them remains compatible.
 module Effectful.Exception
   ( -- * Throwing
     throwIO
--- a/effectful-core/src/Effectful/Exception.hs
+++ b/effectful-core/src/Effectful/Exception.hs
@@ -1,79 +1,85 @@
--- | The 'Eff' monad comes with instances of 'MonadThrow', 'MonadCatch' and
--- 'MonadMask' from the
--- [@exceptions@](https://hackage.haskell.org/package/exceptions) library, so
--- this module simply re-exports the interface of the
--- [@safe-exceptions@](https://hackage.haskell.org/package/safe-exceptions)
--- library.
+{-# LANGUAGE CPP #-}
+-- | Support for runtime exceptions.
 --
--- Why @safe-exceptions@ and not @exceptions@? Because the former makes it much
--- easier to correctly deal with asynchronous exceptions (for more information
--- see its [README](https://github.com/fpco/safe-exceptions#readme)) and
--- provides more convenience functions.
+-- The 'Eff' monad provides instances of 'C.MonadThrow', 'C.MonadCatch' and
+-- 'C.MonadMask', so any function that works with them can be used
+-- seamlessly. This includes API of the following libraries:
+--
+-- - [@exceptions@](https://hackage.haskell.org/package/exceptions)
+--
+-- - [@safe-exceptions@](https://hackage.haskell.org/package/safe-exceptions)
+--
+-- - [@annotated-exception@](https://hackage.haskell.org/package/annotated-exception)
+--
+-- In addition, this module provides thin wrappers over functions from
+-- "Control.Exception" as well as several utility functions for convenience.
 module Effectful.Exception
   ( -- * Throwing
-    C.MonadThrow(..)
-  , Safe.throwString
-  , Safe.StringException(..)
+    throwIO
 
     -- * Catching (with recovery)
-  , C.MonadCatch(..)
-  , Safe.catchIO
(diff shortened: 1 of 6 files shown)


effectful-core/src/Effectful/Exception.hs
```diff
@@ -1,87 +1,112 @@
...
--- Why @safe-exceptions@ and not @exceptions@? Because the former makes it much
--- easier to correctly deal with asynchronous exceptions (for more information
--- see its [README](https://github.com/fpco/safe-exceptions#readme)) and
--- provides more convenience functions.
+{-# LANGUAGE CPP #-}
+-- | Support for runtime exceptions.
+--
+-- The 'Eff' monad provides instances of 'C.MonadThrow', 'C.MonadCatch' and
+-- 'C.MonadMask', so any function that works with them can be used
+-- seamlessly. This includes API of the following libraries:
```
mmhat, line 6:
```suggestion
-- seamlessly. This modules provides an API similar to those of the following libraries:
```

"includes" might be a bit to strong here; Maybe we should go with a phrasing that is a bit more fuzzier.

effectful-core/src/Effectful/Exception.hs
```diff
@@ -113,18 +138,436 @@ module Effectful.Exception
...
+-- :}
+--
+-- >>> isSyncException AsyncEx
+-- Nothing
+--
+-- >>> isAsyncException AsyncEx
+-- Just AsyncEx
+
+-- | Return 'Just' if the given exception is considered synchronous.
+isSyncException :: E.Exception e => e -> Maybe e
```
mmhat, line 528: Hm, shouldn't this be called `maybeSyncException`? I usually expect functions prefixed with "is" to be a predicate; I.e. to return a `Bool`.

effectful-core/src/Effectful/Exception.hs
```diff
@@ -113,18 +138,436 @@ module Effectful.Exception
...
+isSyncException e = case E.fromException (E.toException e) of
+  Just E.SomeAsyncException{} -> Nothing
+  Nothing -> Just e
+
+-- | Return 'Just' if the given exception is considered asynchronous.
+isAsyncException :: E.Exception e => e -> Maybe e
```
mmhat, line 534: Same here; See the comment on `isSyncException`.

mmhat (commented):
LGTM; I just left two small comments, but nothing big :+1:

effectful-core/src/Effectful/Exception.hs
ocharles: I agree, though I'd go with `toSyncException`

mmhat: That's even better :+1:

arybczak: I went with option 3 - made `isSyncException` and `isAsyncException` the same type as in `safe-exceptions` and used `catchIf`/`tryIf` for `catchSync`/`trySync` ;) I missed these helpers earlier, that's why I adjusted these functions to fit the `*Just` variant, but that's unnecessary now.

effectful-core/src/Effectful/Exception.hs
arybczak: I shuffled the top section, should be good now.

effectful-core/src/Effectful/Exception.hs
```diff
@@ -1,87 +1,112 @@
...
+-- | Support for runtime exceptions.
+--
+-- This module provides thin wrappers over functions from "Control.Exception" as
+-- well as several utility functions for convenience.
+--
+-- However, the 'Eff' monad provides instances for 'C.MonadThrow',
+-- 'C.MonadCatch' and 'C.MonadMask', so you can also use the following
+-- libraries:
+--
+-- - [@exceptions@](https://hackage.haskell.org/package/exceptions)
```
Kleidukos, line 11: While I see the point of using `safe-` and `annotated-` variants, is there a concrete recommendation that the documentation could make for new adopters of `effectful`? Like:

> Start with using the functions of this module, and if <conditions> then switch to `exceptions`.

What would motivate someone to use the implementations in `exceptions`?

arybczak: Yeah, I removed this, see https://github.com/haskell-effectful/effectful/pull/257#discussion_r1802865352

Kleidukos: Okay, I see!

arybczak:
Thanks everyone for the comments :slightly_smiling_face:

PR #268, opened 2024-11-04
Support good type inference with unique parametrized effects

arybczak:
I'm very much dissatisfied with the type inference story for parametrized effects.

It's either thousands of type applications or `effectful-plugin` and I don't particularly like the unnecessary verbosity that comes with the former and the fact that the latter is, well, a GHC plugin and GHC plugins have bugs and usually need to be adjusted for every new GHC version.

Considering the fact that if you use an effect with parameters, then in the vast majority of cases there will be only one instance of it you're using at a time, I remembered https://github.com/re-xyr/cleff/issues/23 and managed to package the idea into a nice-looking interface, so now for chosen effects you can do this:

```haskell
data Test1 a :: Effect where
  Test1 :: Test1 a m a
makeTheEffect ''Test1

data Test2 a b :: Effect where
  Test2 :: Test2 a b m (a, b)
  AmbTest :: Test2 a b m ()
makeTheEffect ''Test2

test :: (Num n, IsString s, Test1 n <:> es, Test2 s Char <:> es) => Eff es ()
test = do
  _ <- test1
  _ <- test2
  ambTest
```

(or write helper functions by hand as usual, just using `<:>` instead of `:>`) and it compiles without ambiguity errors.

This PR provides helpers for determining all type parameters up to 5, but one can write a local helper class for any combination.

Thoughts about naming? I think `TheN` helpers are fine (I thought about `UniqueN`, but I like `TheN` better). Not quite sure about `<:>`, though I couldn't come up with anything better.

TODO:
- [ ] Better naming?
- [ ] Documentation.
- [ ] Helper functions for relevant built-in effects (both static and dynamic).

### DIFF
--- a/effectful-core/effectful-core.cabal
+++ b/effectful-core/effectful-core.cabal
@@ -1,7 +1,7 @@
 cabal-version:      3.0
 build-type:         Simple
 name:               effectful-core
-version:            2.5.0.0
+version:            2.5.1.0
 license:            BSD-3-Clause
 license-file:       LICENSE
 category:           Control
--- a/effectful-core/src/Effectful.hs
+++ b/effectful-core/src/Effectful.hs
@@ -27,6 +27,7 @@ module Effectful
   , Dispatch(..)
   , DispatchOf
   , (:>)
+  , type (<:>)
   , (:>>)
 
     -- * Running the 'Eff' monad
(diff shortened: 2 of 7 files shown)


mmhat:
I think that would be a fine addition to the library!
One comment about the naming: When I read `<:>` it makes me think that that operator is symmetric - I think it's to close to `<>`.
I propose `!:>`, `:!>` or `!>` (preference on the first): While not quite related to Haskell the exclamation mark reminds me of [uniqueness quantification](https://en.wikipedia.org/wiki/Uniqueness_quantification) as opposed to [existential quantification](https://en.wikipedia.org/wiki/Existential_quantification) used in some logics.

PR #289, opened 2024-12-22
Add Input and Output effects

arybczak:
Proof of concept.

- `Effectful.Input.Const` as a restricted version of `Reader` for ensuring that the input parameter doesn't change (static only, since with a dynamic you can change it with `interpose`).
- `Effectful.Output.Array` is the correct alternative to `Writer` when you just want to accumulate values.

And then it turned out that these are generalized by the `Coroutine` effect, so might as well also include it, it looks like there's no point providing a dynamic `Input` and `Output`.

Opinions welcome.


arybczak:
@tomjaguarpaw perhaps you have an opinion about this, i.e. separate Input/Output vs type synonyms for Coroutine?

tomjaguarpaw:
If I was doing it again I would have made `Stream` and `Produce` newtypes of `Coroutine`, rather than type synonyms. The error messages would have been better. I don't know that the same reasoning holds for `effectful` though.

PR #295, opened 2025-01-31
Add new EarlyReturn effect

kedashoe:
Thank you all again for the help in https://github.com/haskell-effectful/effectful/discussions/285

This PR uses your code block, @arybczak , with the two modifications you suggest as I feel that is the more common use case. Similarly, I gave the "best" name, `runEarlyReturn`, to the handler returning `Eff es a`.

As my further replies in that discussion suggest, I struggled for a bit more getting everything working how I ideally wanted but did get there in the end. The docs I've included are targeted at beginner/intermediate, this sort of thing would have helped me and I figure people who know what they are doing will just glance at it and move on :)

### DIFF
--- a/effectful-core/effectful-core.cabal
+++ b/effectful-core/effectful-core.cabal
@@ -86,6 +86,7 @@ library
                      Effectful.Dispatch.Static
                      Effectful.Dispatch.Static.Primitive
                      Effectful.Dispatch.Static.Unsafe
+                     Effectful.EarlyReturn.Dynamic
                      Effectful.Error.Dynamic
                      Effectful.Error.Static
                      Effectful.Exception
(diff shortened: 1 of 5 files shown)


kedashoe:
Forgot about this one, I'll go ahead and close, I imagine this approach is not your cup of tea? I do think though it would be great to get some form of early return into effectful! Thanks again for the help in the discussion.

ocharles:
I don't think you should close this - it looks good to me! We would benefit from this at work. I'd just nudge @arybczak and get more clarification

ocharles:
@arybczak what are your thoughts on this?

arybczak:
The effect is fine, I'm just occupied with other things at the moment and didn't put much attention to non-trivial feature work for this project.

I'll most likely resurrect it at some point before the 2.6.0.0 release. I need to deal with #316 first.

PR #307, opened 2025-03-06
Add withException, similar to safe-exceptions and unliftio. Closes #304

### LINKED ISSUE #304: Add something like Control.Exception.Safe.withException to Effectful.Exception.
The documentation says:

> If you want to catch all exceptions, run a cleanup action and rethrow, use one of the functions from the cleanup section.

While several of those functions let you specify a an action to be performed when an exception is thrown, they don't give you access to the actual exception. This problem is shared with the Control.Exception module in the base package, but the safe-exceptions package contains this useful function that otherwise behaves like onException:

> withException :: Exception e => IO a -> (e -> IO b) -> IO a

It would be great if effectful could provide a similar function, since it's easy to get the edge cases wrong when dealing with asynchronous exceptions.

torgeirsh:
I copied and edited the [code from unliftio](https://hackage.haskell.org/package/unliftio-0.2.25.0/docs/src/UnliftIO.Exception.html#withException). Since other functions from Effectful.Exception are used, the withRunInIO in the original code isn't needed, but I don't know if it would be better to use reallyUnsafeUnliftIO and call the Control.Exception functions directly. I'm not an expert on exceptions, so I don't know if it affects the safety.

For reference, [safe-exceptions has a slightly different implementation](https://hackage.haskell.org/package/safe-exceptions-0.1.7.4/docs/src/Control.Exception.Safe.html#withException) that uses generalBracket.

### DIFF
--- a/effectful-core/CHANGELOG.md
+++ b/effectful-core/CHANGELOG.md
@@ -1,4 +1,5 @@
 # effectful-core-2.6.0.0 (????-??-??)
+* Add `withException` to `Effectful.Exception`, similar to safe-exceptions and unliftio.
 * **Breaking changes**:
   - Change the order of type parameters in `raise` for better usability.
   - `Effectful.Error.Static.ErrorWrapper` is no longer caught by `catchSync`.
--- a/effectful-core/src/Effectful/Exception.hs
+++ b/effectful-core/src/Effectful/Exception.hs
@@ -59,6 +59,7 @@ module Effectful.Exception
   , C.ExitCase(..)
   , finally
   , onException
+  , withException
 
     -- * Utils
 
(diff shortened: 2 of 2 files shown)


arybczak:
Superseded by #311.

torgeirsh:
Thanks! I noticed that the replacement doesn't use masking. Isn't there a risk of losing the original exception if a new asynchronous exception occurs between the catch and the rethrow?

arybczak:
> the replacement doesn't use masking

It does, `catch` masks the exception handler. It uses `mask` rather than `uninterruptibleMask`, but that's a tradeoff that's discussed [here](https://github.com/fpco/safe-exceptions/issues/3) if you're interested in details.

`Effectful.Exception` is supposed to be a thin wrapper for `Control.Exception` so it doesn't use `uninterruptibleMask`, but if someone really wants the behavior of `safe-exceptions`, the library works just fine with `effectful` ;)

PR #317, opened 2025-04-05
Describe laws for NonDet

michaelpj:


arybczak:
Thanks. Can you please retain the single-line comment style?

torgeirsh:
Would it be accurate to say that NonDet essentially is the Effectful equivalent to MaybeT?

arybczak:
@torgeirsh Yes, indeed. It looks like it would be good to make a section about this (i.e. what effects correspond to what transformers) in haddocks of the main `Effectful` module.

PR #339, opened 2025-08-10
Add Effectful.ST and Effectful.ST.STRef

Innf107:
With this, effectful is able to mix `STE` with other effects, which makes it a very convenient way to avoid having to depend on `IOE` just to locally use mutable data structures.

The effect is tagged with an `s` region parameter and `runSTE` uses the same higher rank type trick as regular `runST` to prevent references from escaping.

I'm pretty sure that this is sound, however `ST` can be subtle so it would be nice to have a second opinion.
For some prior art: Koka uses [a similar `st` effect](https://koka-lang.github.io/koka/doc/std_core_types.html#type_space_st) that can be combined with arbitrary other effects.
References escaping through other effects (e.g. exceptions) shouldn't be an issue since that can only happen through an existential, at which point the reference cannot be accessed anymore anyway.

One thing I'm not entirely sure about is the type role on `STE`. The region parameter needs to be nominal for soundness, but making the others `phantom` seems a bit dangerous? This is also what is inferred for other effects like `Process` though.

### DIFF
--- a/effectful/CHANGELOG.md
+++ b/effectful/CHANGELOG.md
@@ -3,6 +3,7 @@
   `Eff` for compatibility with existing code.
 * Re-export `writeTMVar` from `stm-2.5.1.0` in `Effectful.Concurrent.STM`.
 * Add `cancelMany` to `Effectful.Concurrent.Async`.
+* Add `Effectful.ST` and `Effectful.ST.STRef`
 
 # effectful-core-2.6.0.0 (2025-06-13)
 * Adjust `generalBracket` with `base >= 4.21` to make use of the new exception
--- a/effectful/effectful.cabal
+++ b/effectful/effectful.cabal
@@ -106,6 +106,8 @@ library
                      Effectful.Prim.IORef
                      Effectful.Prim.IORef.Strict
                      Effectful.Process
+                     Effectful.ST
+                     Effectful.ST.STRef
                      Effectful.Temporary
                      Effectful.Timeout
 
@@ -171,6 +173,7 @@ test-suite test
                     PrimTests
                     ReaderTests
                     StateTests
+                    STTests
                     TimeoutTests
                     UnliftTests
                     Utils
--- a/effectful/src/Effectful/ST.hs
+++ b/effectful/src/Effectful/ST.hs
@@ -1,0 +1,28 @@
+module Effectful.ST (STE, runSTE, liftST) where
+
+import Control.Monad.ST (ST)
+import Control.Monad.ST.Unsafe (unsafeSTToIO)
+import Data.Kind (Type)
+import Data.Proxy (Proxy (..))
+
+import Effectful
+import Effectful.Dispatch.Static
+
+-- | An effect for embedding 'ST' computations
+data STE (s :: Type) :: Effect
+type role STE nominal phantom phantom
+
+type instance DispatchOf (STE s) = Static NoSideEffects
+data instance StaticRep (STE s) = STE
+
+-- | Run the 'STE' effect.
+-- Since effectful allows several 'STE' effects to be in scope at once,
+-- it can be ambiguous which one a usage of 'liftST' refers to.
+-- In these cases, the 'Proxy' parameter can be used to disambiguate.
+-- In particular, instead of @liftST $ newSTRef x@, you will have to write 
+-- @liftST \@s $ newSTRef x@.
+runSTE :: (forall s. Proxy s -> Eff (STE s : es) a) -> Eff es a
+runSTE eff = evalStaticRep STE (eff Proxy)
+
+liftST :: forall s a es. (STE s :> es) => ST s a -> Eff es a
+liftST st = unsafeEff_ (unsafeSTToIO st)
--- a/effectful/src/Effectful/ST/STRef.hs
+++ b/effectful/src/Effectful/ST/STRef.hs
@@ -1,0 +1,28 @@
+-- | Lifted "Data.STRef"
+module Effectful.ST.STRef (STRef, newSTRef, readSTRef, writeSTRef, modifySTRef, modifySTRef') where
+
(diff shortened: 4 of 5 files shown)


arybczak:
Thanks for the PR :+1: However...

Did you see https://hackage-content.haskell.org/package/effectful-core-2.6.0.0/docs/Effectful-Prim.html? It allows you to use everything from `primitive`, which pretty much generalizes and extends `ST`. The only thing you lose is the scope restriction (and thus `runPrim` requres `IOE`, so you can't use it within `runPureEff`), but it's IMO the best compromise.

I considered making it `Prim s` but the type parameter doesn't play well with type inference (as you already noticed since you're talking about having to use type applications and introduced a proxy). It really requires effectful-plugin in order to not be very annoying to use.

For example code like this:

```
blub :: STE s :> es => Eff es ()
blub = pure ()
```

won't compile without `effectful-plugin` if you don't have `AllowAmbiguousTypes` enabled.

In conclusion: I experimented with this approach and decided it's not worth it over how currently `Prim` operates, so I'm not going to include it in the core library. But it's of course a great candidate for a separate library.

Innf107:
> Did you see https://hackage-content.haskell.org/package/effectful-core-2.6.0.0/docs/Effectful-Prim.html? 

I have! `Prim` is very nice for giving functions the ability to write to shared mutable references without exposing `IO`, but the problem compared to `STE` is that `Prim` still leaks to consumers. It's more like an `IORef` effect than `ST` per se.

The main use case I see for `ST(E)` is as an implementation detail of functions that may otherwise not depend on Prim.

For example, what motivated me to make this PR was a very generic graph algorithm of type (essentially)
```hs
computeSCC :: (Hashable node) => (node -> Eff es [node]) -> node -> Eff es (HashMap node SCCId)
```
The function to access a node's edges needs to run in an effect since edges sometimes need to be computed from a mutable structure, so I can't just use regular ST, but I also don't want to force Prim onto callers since some of them may already have the full immutable graph available and use `runPureEff`.

As for the ambiguity, I agree that it is a bit unfortunate, but I don't think it's as bad as it seems.

A function using `STE` is only ambiguous, if it doesn't mention the passed in region in its arguments or return type. But in that case, there is really no point in requiring a region to be passed in over just running a new one.

So an example like your `blub` could always be rewritten unambiguously like this with no downside.
```hs
blub :: Eff es ()
blub = runSTE \(Proxy :: Proxy s) -> do
    _ <- newSTRef @s 'x'
    pure ()
```

Moreover, the only operations that ever need to use the `s` parameter are operations that create a new `s` region. Any `read/write` operation or anything else that takes an existing object parameterized on `s` (like an `STRef` or hash table) is already unambiguous.

So really, the only operations that need type applications are operations like `newSTRef` that create new objects, which are going to be comparatively rare anyway. I think that's pretty managable.

Innf107:
As a concrete example, the SCC algorithm I mentioned works with `STE` with no type annotations aside from the parts that create` STRef`s/hashtables and otherwise needed no modifications going from `IORef`s to `STRef`s other than changing the names of all the functions and adding rasie before `outEdgesOrPrecomputedSCC`


<details>
<summary>(simplified to reduce dependencies so i could check it in the effectful repo)</summary>

```hs
computeSCC ::
    forall node es.
    (Show node, Ord node, IOE :> es) =>
    (node -> Eff es (Maybe [node])) ->
    node ->
    Eff es (Map node Int)
computeSCC outEdgesOrPrecomputedSCC node = runSTE $ \(Proxy :: Proxy s) -> do
    visited :: STRef s (Set node) <- newSTRef [node]
    currentDFSNum :: STRef s Int <- newSTRef 0
    dfsNums :: STRef s (Map node Int) <- newSTRef mempty

    openSCCs :: STRef s [node] <- newSTRef [node]
    openNodes :: STRef s [node] <- newSTRef [node]

    sccs :: STRef s (Map node Int) <- newSTRef mempty

    sccIds :: STRef s Int <- newSTRef 0
    let newSCCId = do
            id <- readSTRef sccIds
            writeSTRef sccIds (id + 1)
            pure id

    let go node = do
            dfsNum <- readSTRef currentDFSNum
            writeSTRef currentDFSNum (dfsNum + 1)
            modifySTRef' dfsNums (Map.insert node dfsNum)
            raise (outEdgesOrPrecomputedSCC node) >>= \case
                Nothing -> pure ()
                Just neighbors -> do
                    for_ neighbors $ \neighbor -> do
                        visitedUntilNow <- readSTRef visited
                        case Set.member neighbor visitedUntilNow of
                            False -> do
                                modifySTRef' visited (Set.insert neighbor)
                                modifySTRef' openSCCs (neighbor :)
                                modifySTRef' openNodes (neighbor :)

                                go neighbor
                            True -> do
                                dfsNumsUntilNow <- readSTRef dfsNums
                                let dfsNumOf otherNode = case Map.lookup otherNode dfsNumsUntilNow of
                                        Nothing -> error $ "DFS number of '" <> show node <> "' not found"
                                        Just dfsNum -> dfsNum

                                modifySTRef' openSCCs (dropWhile (\representative -> dfsNumOf representative >= dfsNum))
                    -- backtrack
                    readSTRef openSCCs >>= \case
                        (topRepresentative : rest)
                            | topRepresentative == node -> do
                                writeSTRef openSCCs rest

                                sccId <- newSCCId

                                let assignSCC = \case
                                        [] -> pure []
                                        (openNode : rest) -> do
                                            modifySTRef' sccs (Map.insert openNode sccId)
                                            if openNode == node
                                                then pure rest
                                                else assignSCC rest
                                currentOpenNodes <- readSTRef openNodes
                                remainingOpenNodes <- assignSCC currentOpenNodes
                                writeSTRef openNodes remainingOpenNodes
                        _ -> pure ()

    go node
    readSTRef sccs

```
</details>

Issue #353, opened 2026-04-10
Lazier initialisation of effectful-plugin?


torgeirsh:
In a Cabal project with multiple executables, it can be convenient to add -fplugin=Effectful.Plugin to the common GHC options shared by all of them. If one of the executables doesn't use Effectful, that results in an error:

> \<no location info\>: error:
>     Please add effectful-core to the list of dependencies.

If I do that, I get a warning:

> \<no location info\>: warning: [GHC-42258] [-Wunused-packages]
>     The following packages were specified via -package or -package-id flags,
>     but were not needed for compilation:
>       - effectful-core-2.6.1.0 (exposed by flag -package-id effectful-core-2.6.1.0-6tso3GaPoudRJxw7iWIoZ)

I can of course ignore the warning, or work around it by only adding -fplugin=Effectful.Plugin to the executables that use it, but it would be convenient if the plugin didn't complain about the missing effectful-core dependency until it encountered code that uses Effectful. Is lazy initialisation possible, or is it a chicken-and-egg problem where the plugin needs effectful-core in order to determine that there's code that uses Effectful?

arybczak:
> is it a chicken-and-egg problem where the plugin needs effectful-core in order to determine that there's code that uses Effectful

That's exactly it, so unfortunately what you're asking doesn't seem possible.

torgeirsh:
@arybczak There's a related issue when adding the plugin to a component that only contains imports from the effectful package, e.g. Effectful.Concurrent.Async. An effectful-core dependency must be added for the plugin, but whether I get a warning about effectful-core being unused seems to depend on whether the code contains any expressions that must be disambiguated by the plugin, which can change over time and feels a bit arbitrary. I can add a dummy import Effectful () to ensure that effectful-core is always considered used, but could the plugin ensure that instead, since it already requires effectful-core?

PR #362, opened 2026-06-17
Inject empty Effectful import into modules the plugin runs in

arybczak:
@torgeirsh this should fix your problem from https://github.com/haskell-effectful/effectful/issues/353#issuecomment-4266756442.


arybczak:
Heh, it actually doesn't work, i.e. -Wunused-packages still reports that effectful-core is not used even though it's needed for compilation. Looks like a bug in the warning flag (not the only one).

PR #365, opened 2026-07-14
Add OsPath variants of the FilePath-based APIs

### LINKED ISSUE #329: Integrating OsPath APIs in effectful
@arybczak I have been prototyping with having `OsPath`-base APIs in Effectful. Today we have the following:

* Effectful.FileSystem
  * Effectful.FileSystem.IO
    * Effectful.FileSystem.IO.ByteString
       * Effectful.FileSystem.IO.ByteString.Builder
       * Effectful.FileSystem.IO.ByteString.Lazy
     * Effectful.FileSystem.IO.File

My first attempt was to add `*.OsPath` modules, giving us:

* Effectful.FileSystem
  * **Effectful.FileSystem.OsPath**
  * Effectful.FileSystem.IO
  * **Effectful.FileSystem.OsPath**
    * Effectful.FileSystem.IO.ByteString
       * **Effectful.FileSystem.IO.ByteString.OsPath**
       * Effectful.FileSystem.IO.ByteString.Builder
       * Effectful.FileSystem.IO.ByteString.Lazy
         * **Effectful.FileSystem.IO.ByteString.Lazy.OsPath**
     * Effectful.FileSystem.IO.File
       *  **Effectful.FileSystem.IO.File.OsPath**

From what I understand, we have two dimensions that determine the amount of APIs:

* Does the API take a `FilePath` or an `OsString`?
* Does the API return a `StrictByteString` or a `LazyByteString`?

There is probably the opportunity to bundle some of these modules together? I'm having a hard time understanding why the `.IO.File` module does not have actions on file like `removeFile`, for instance.

In particular I'd like to challenge the existence of the `.IO` module hierarchy, since there the effect is static and thus cannot have a pure interpreter.

If you're down with it I can make suggestions to redesign the module hierarchy so that we have alternative APIs that don't duplicate that many modules.

Kleidukos:
I'm more and more in need of handling OsPath values within my applications, and ceaselessly manually converting back and forth between FilePath and OsPath is unwieldy. 

@arybczak Let me know what you think of the module layout.

Closes #329

### DIFF
--- a/effectful/CHANGELOG.md
+++ b/effectful/CHANGELOG.md
@@ -1,4 +1,8 @@
 # effectful-2.7.0.0 (2026-??-??)
+* Add `OsPath` variants of the `FilePath` based APIs
+  (`Effectful.FileSystem.OsPath` for `System.Directory.OsPath` and
+  `Effectful.FileSystem.File.OsPath` for `System.File.OsPath` from the
+  `file-io` package).
 * Add the `Input` effect (`Effectful.Input.Dynamic`, `Effectful.Input.Static`,
   `Effectful.Input.Static.Action` and `Effectful.Labeled.Input`) for access to
   values.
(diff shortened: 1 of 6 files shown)


arybczak:
Looks good, thanks :+1:

PR #381, opened 2026-09-09
Run pure computations in a separate thread

### LINKED ISSUE #380: `runPureEff` values are permanently poisoned by asynchronous exceptions delivered to the thread forcing them
A thunk whose evaluation runs `runPureEff` with a static effect handler such as `evalState` is not safe to share between threads. If the thread forcing it receives an asynchronous exception mid-evaluation, the thunk is overwritten with a closure that re-raises that exception. That means that every other consumer of the value _also_ gets the exception. A plain pure thunk in the same situation is instead resumed when another consumer tries to force it.

To really be hit by this you have to observe the thunk from another thread. But that's reasonably easy in practice:
- Worker thread T1 runs a pure computation C using `runPureEff` and stores it in a TVar as a thunk (i.e. does not force it)
- Other threads T2, T3 read the TVar
- T2 is killed while forcing C
- T3 evaluates C and gets the exception that killed T2!

This means that you can't safely use an `effectful` pure `State` loop as a drop-in replacement for e.g. `StateT`.

# Analysis

`runPureEff` uses `unsafeDupablePerformIO`. The static handlers run under `inlineBracket`, so evaluation of the thunk has a catch frame
on the stack. When an asynchronous exception arrives, the RTS delivers it to that handler; the handler rethrows with `throwIO`, which is a _synchronous_ raise. A synchronous raise walks the remaining stack and updates every thunk under evaluation with a "raise this exception" closure. Without the catch frame the RTS would instead freeze the evaluation into a resumable `AP_STACK`.

The base documentation for `unsafeDupablePerformIO` already says that `bracket` cannot be used safely inside it. The comment on `runPureEff` argues it is safe because no side effects are performed, but sadly it seems it does change the exception behaviour.

# Reproducer

Reproducer attached. This shows the expected behaviour for pure computations with a plain thunk and `StateT`, with `runPureEff` and a simplified `unsafeDupablePerformIO` version showing the bad behaviour.

```
plain pure thunk:
  worker:              raised thread killed
  later force by main: ok
transformers evalState thunk:
  worker:              raised thread killed
  later force by main: ok
unsafeDupablePerformIO + bracket thunk:
  worker:              raised thread killed
  later force by main: raised thread killed
runPureEff + evalState thunk:
  worker:              raised thread killed
  later force by main: raised thread killed
```

# Possible solutions

- Document the problems with `runPureEff`. I don't like this as the problem can occur arbitrarily far from where `runPureEff` is used - you can bury it deep in a nominally pure library.
- Avoid catch frames in the pure interpreter: `inlineBracket` exists to restore the environment on exception, but in `runPureEff` there is no caller to observe the environment afterwards, so maybe we could somehow avoid the bracket.

-----

```
{-# LANGUAGE BangPatterns #-}
{-# LANGUAGE BlockArguments #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TypeApplications #-}
module Main (main) where

import Control.Concurrent
import Control.Exception
import Control.Monad
import Data.IORef
import Control.Monad.Trans.State.Strict qualified as T
import Effectful (runPureEff)
import Effectful.State.Static.Local (evalState, get, put)
import System.IO
import System.IO.Unsafe (unsafeDupablePerformIO)

-- A "pure" computation that takes a while: sum 1..n through effectful's State.
pureSum :: Int -> Int
pureSum n = runPureEff $ evalState (0 :: Int) do
  forM_ [1 .. n] \i -> do
    s <- get
    put $! s + i
  get
{-# NOINLINE pureSum #-}

-- The same computation through transformers' State, for comparison.
transformersSum :: Int -> Integer
transformersSum n = flip T.evalState (0 :: Integer) do
  forM_ [1 .. n] \i -> do
    s <- T.get
    T.put $! s + fromIntegral i
  T.get
{-# NOINLINE transformersSum #-}

-- The same computation with no monad at all, for comparison.
plainSum :: Int -> Integer
plainSum n = go 0 1
  where
    go !acc i
      | i > n = acc
      | otherwise = go (acc + fromIntegral i) (i + 1)
{-# NOINLINE plainSum #-}

-- No effectful, but the same shape as runPureEff + evalState: unsafeDupablePerformIO
-- around a bracket (effectful's runState is unsafeDupablePerformIO around inlineBracket).
bracketSum :: Int -> Int
bracketSum n = unsafeDupablePerformIO do
  ref <- newIORef 0
  bracket_ (pure ()) (pure ()) do
    forM_ [1 .. n] \i -> modifyIORef' ref (+ i)
  readIORef ref
{-# NOINLINE bracketSum #-}

-- Force a shared thunk in a worker thread, kill the worker while it is mid-evaluation,
-- then force the same thunk again from the main thread and report what happens.
scenario :: String -> a -> IO ()
scenario label thunk = do
  shared <- newIORef thunk
  done <- newEmptyMVar
  worker <- forkIO do
    r <- try @SomeException (readIORef shared >>= evaluate)
    putMVar done r
  threadDelay 20_000 -- let the worker get going
  killThread worker
  workerResult <- takeMVar done
  later <- try @SomeException (readIORef shared >>= evaluate)
  putStrLn $ label <> ":"
  putStrLn $ "  worker:              " <> either (("raised " <>) . show) (const "finished") workerResult
  putStrLn $ "  later force by main: " <> either (("raised " <>) . show) (const "ok") later

main :: IO ()
main = do
  hSetBuffering stdout LineBuffering
  let n = 200_000_000
  scenario "plain pure thunk" (plainSum n)
  scenario "transformers evalState thunk" (transformersSum n)
  scenario "unsafeDupablePerformIO + bracket thunk" (bracketSum n)
  scenario "runPureEff + evalState thunk" (pureSum n)
```

arybczak:
A thunk that runs runPureEff with an effect handler was not safe to share between threads. Effect handlers install catch frames, so an asynchronous exception delivered to the thread forcing the thunk was rethrown synchronously, which overwrote the thunk with a closure that reraises it. Every other consumer of the value then got that exception, while a plain pure thunk is resumed instead.

Run the computation in a forked thread and wait for its result, so that the stack of the thread forcing the thunk holds no catch frame and the RTS suspends the evaluation into a resumable AP_STACK.

Fixes #380.


arybczak:
```haskell
runPureEff :: HasCallStack => Eff '[] a -> a
runPureEff (Eff m) = do
  unsafeDupablePerformIO $ do
    let go = (m =<< emptyEnv) `E.catchNoPropagate`
          \(E.ExceptionWithContext _ (e::E.SomeException)) -> do
            tid <- myThreadId
            E.throwTo tid e
            go
    go
```

is a much cheaper fix, but it doesn't fully work because of https://gitlab.haskell.org/ghc/ghc/-/work_items/27794.

